PR #3, opened 2015-05-05
un-html, formatting, adding text from discussions and drafts

dericed:
@epiil and I started to integrate text from email discussions to the specification. I also did a lot of work to remove html formatting and other formatting that unlike other RFC documents. Because these patches integrate some of the discussion and draft with the specification there are now a few internal contradictions (such as the comments about the discrepancies in notes about the possible number of Element IDs). I can file issues for those or continue to discuss on the listserv.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -35,10 +35,12 @@ Master-element   | contains other EBML sub-elements of the next lower
                  | level
 Binary           | not interpreted by the parser
 
-As well as defining standard data types, EBML uses a system of Elements
-to make up an EBML "document". Elements incorporate an Element ID, a
-descriptor for the size of the element, and the binary data itself.
-Futher, Elements can be nested or contain Elements of a lower "level".
+## Structure
+
+EBML uses a system of Elements to compose an EBML "document". Elements
+incorporate three parts: an Element ID, an Element Data Size, and
+Element Data. The Element Data, which is described by the Element ID, 
+may include either binary data or one or many other Elements.
 
 Element IDs are outlined as follows, beginning with the ID itself,
 followed by the Data Size, and then the non-interpreted Binary itself:


specification.markdown
```diff
@@ -97,242 +170,298 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+encoding. The variable size integer begins with zero or more zero bits
+to define the width of the integer. Zero zeroes means a width of one
+byte, one zero a width of two bytes etc. The zeroes are followed by a
+marker of one set bit and then follows the actual integer value. The
+integer data consists of alignment data and tail data. The alignment
+data together with the width descriptor and the marker makes up one ore
+more complete bytes. The tail data is as many bytes as there were
+zeroes in the width descriptor, i.e. width-1.
+
+VINT           = VINT_WIDTH VINT_MARKER VINT_DATA
```
robUx4: There's a formatting issue on the markdown rendering

specification.markdown
```diff
@@ -97,242 +170,298 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
- </tbody>
-</table>
+## Structure
+
+Syntactically an EBML document is a list of EBML elements. Each element
+has three parts; an Element ID, an Element Data Size and the element
+data payload. The element data payload is then either data, implicitly
+typed by the Element ID, or a list of EBML elements.
+
+EBML       = *ELEMENT
```
robUx4: There's a formatting issue on the Markdown rendering

specification.markdown
```diff
@@ -97,242 +170,298 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+ELEMENT_ID = VINT
+SIZE       = VINT
+
+EBML uses big endian/network order byte order, i.e. most significant
+bit first. All of the tokens above are byte aligned.
+
+## Variable Size Integer
+
+For both Element ID and Element Data Size EBML uses a variable size
+integer, coded according to a schema similar to that of UTF-8 [UTF-8]
```
robUx4: the `[UTF-8]` should probably link somewhere or is redundant ?

specification.markdown
```diff
@@ -97,242 +170,298 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+    Multiple:       No
+    Range:          -
+    Default:        -
+    Element Type:   Binary
+    Description:    Used to void damaged data, to avoid unexpected
+                    behaviors when using damaged data. The content is
+                    discarded. Also used to reserve space in a
+                    sub-element for later use.
+
+### signature
```
robUx4: Should mention that it's deprecated.

robUx4:
As a general remark, avoid more than one commit per pull request. Git tools allow to turn several commits into one. You'll have a harder time editing your pull request when you have to edit the 1st of a batch of 37 commits.

Since it's the early work and a lot of changes are needed it's fine. The detail of why a change at a specific line/word won't matter much.

specification.markdown
dericed: Cleaned up in https://github.com/MediaArea/ebml-specification/commit/ca14fd2b9ec38cc4cae9b8730ff89a3f97160e6c

specification.markdown
dericed: fixed in https://github.com/MediaArea/ebml-specification/commit/a3ae17f4d8e77295ff899a2b7a4a0923c6105e12

specification.markdown
dericed: ah, I misunderstood what you were talking about at first. New patch to format the list: https://github.com/MediaArea/ebml-specification/commit/a3ae17f4d8e77295ff899a2b7a4a0923c6105e12

specification.markdown
dericed: @robUx4 Is the signature section deprecated or simply in draft form and never finalized?

dericed:
any further pending issues on this PR?

specification.markdown
robUx4: I'd say in deprecated. libebml doesn't handle it. So not sure if anyone ever implemented support.
A signature is good if it's external to the content. Or, as in PGP, to prove the writer is the write person, but then it requires some key exchange. To verify data are not altered we already have CRC. Beyond that it's overkill with no added security.

specification.markdown
```diff
@@ -1,66 +1,140 @@
...
+
+This document describes the EBML binary syntax, its semantic
+interpretation and the syntax of a textual document type definition
+format used to define the rules and constraints of an EBML language.
+
+## Notations and Conventions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in RFC 2119.
```
robUx4: Link to the RFC needed

mbunkus:
@dericed Are you continuing to squash commits? Or should we merge this?

specification.markdown
```diff
@@ -1,66 +1,140 @@
...
-    precise beginning of the millennium (at
-    2001-01-01T00:00:00,000000000 UTC)
--   master-element - contains other EBML sub-elements of the next lower
-    level
--   Binary - not interpreted by the parser
+Data Type        | Definition
+-----------------|----------------------------------------------------
+Signed Integer   | Big-endian, any size from 1 to 8 octets
+Unsigned Integer | Big-endian, any size from 1 to 8 octets
+Float            | Big-endian, defined for 4 and 8 octets (32, 64
```
robUx4: Do we need the line formatting here ? We could at least have a 80 chars policy. This one ends at 66.

mbunkus: I slightly prefer a tabular layout as it seems to be a bit easier to read, but I do agree that the lines are awfully short.

specification.markdown
```diff
@@ -1,66 +1,140 @@
...
+Binary           | not interpreted by the parser
 
 As well as defining standard data types, EBML uses a system of Elements
-to make up an EBML "document." Elements incorporate an Element ID, a
-descriptor for the size of the element, and the binary data itself.
-Futher, Elements can be nested, or contain, Elements of a lower "level."
+to make up an EBML "document". Elements incorporate an Element ID, a
+descriptor for the size of the element (known as the Element Data
+Size), and the binary data itself. Futher, Elements can be nested or
+contain Elements of a lower "level".
```
robUx4: I think the "lower" level is confusing, given in the specs we increment the number for nested elements. But that can be another pull request.

specification.markdown
```diff
@@ -1,66 +1,140 @@
...
+default, widths up to 4. Another maximum width value can be set by
+setting another value to EBMLMaxIDLength in the EBML header. See
+section 5.1. Element IDs are always encoded in their shortest form, e.g. 1 is always encoded as 0x81 and never as 0x4001. This limits the number of Element IDs in every class with the number of Element IDs in the previous classes. Furthermore, values with all data bits set to 1 and all data bits set to 0 are reserved, hence the effective number of Element IDs are as follows for different widths. Note that the shortest encoding form for 127 is 0x407f since 0x7f is reserved.
+
+Width | Class ID | Number of Element IDs
+------|----------|------------------------
+1     | Class A  | 2^7-2     = 126
+2     | Class B  | 2^14-2^7  = 16,256
+3     | Class C  | 2^21-2^14 = 2,080,768
+4     | Class D  | 2^28-2^21 = 266,338,304
```
mbunkus: I just noticed this. We were talking about »Class-ID« and »Class ID« and »Element ID« and that all those terms are confusing. I don't see a particular need for »Class ID« here and would prefer »Class name« as the column header. That would cut down on the number of uses of »xyz ID«.

specification.markdown
```diff
@@ -1,66 +1,140 @@
...
+
 
-Element IDs are outlined as follows, beginning with the ID itself,
-followed by the Data Size, and then the non-interpreted Binary itself:
+### Element ID
+
+The EBML Element ID is encoded as a variable size integer with, by
+default, widths up to 4. Another maximum width value can be set by
+setting another value to EBMLMaxIDLength in the EBML header. See
+section 5.1. Element IDs are always encoded in their shortest form, e.g. 1 is always encoded as 0x81 and never as 0x4001. This limits the number of Element IDs in every class with the number of Element IDs in the previous classes. Furthermore, values with all data bits set to 1 and all data bits set to 0 are reserved, hence the effective number of Element IDs are as follows for different widths. Note that the shortest encoding form for 127 is 0x407f since 0x7f is reserved.
```
robUx4: This limitation to the shortest form was not in the original specs. I think libebml doesn't have this limitation.
This should be part of another pull request as it's subject to debate.

specification.markdown
```diff
@@ -1,66 +1,140 @@
...
-        Length Descriptor.
-    -   Any ID where all x's are composed entirely of 1's is a Reserved
-        ID, thus the -1 in the definitions above.
-    -   The Reserved IDs (all x set to 1) are the only IDs that may
-        change the Length Descriptor.
+        length of the Element ID. The number of leading 0's + 1 is the
+        length of the ID in octets. We will refer to the leading bits
+        as the Length Descriptor.
+    -   Any Element ID where all x's are composed entirely of 1's is a
+        Reserved Element ID, thus the -1 in the definitions above.
```
robUx4: It's actually -2 above.

specification.markdown
mbunkus: It's coming from the RFC; neither the old EBML specs nor the Matroska specs contain such a limiation. So yes, we should move that to a separate pull request, please, as backwards compatibility should be considered.

robUx4:
In general this pull request mixes formatting and actually spec changes/additions/clarifications.

I think we should go first with formatting only. Then we can discuss the actual wording of things. In general cosmetic changes should never be mixed with content changes.

specification.markdown
```diff
@@ -97,242 +170,298 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+with the width descriptor and the marker makes up one or more complete
+bytes. The tail data is as many bytes as there were zeroes in the width
+descriptor, i.e. width-1.
+
+    VINT           = VINT_WIDTH VINT_MARKER VINT_DATA
+    VINT_WIDTH     = *%b0
+    VINT_MARKER    = %b1
+    VINT_DATA      = VINT_ALIGNMENT VINT_TAIL
+    VINT_ALIGNMENT = *BIT
+    VINT_TAIL      = *BYTE
```
robUx4: How can this be a BYTE when the size is 1 octet ?

specification.markdown
```diff
@@ -97,242 +170,298 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+consists of alignment data and tail data. The alignment data together
+with the width descriptor and the marker makes up one or more complete
+bytes. The tail data is as many bytes as there were zeroes in the width
+descriptor, i.e. width-1.
+
+    VINT           = VINT_WIDTH VINT_MARKER VINT_DATA
+    VINT_WIDTH     = *%b0
+    VINT_MARKER    = %b1
+    VINT_DATA      = VINT_ALIGNMENT VINT_TAIL
+    VINT_ALIGNMENT = *BIT
```
robUx4: Isn't the alignment (padding?) just `*%b0` ?

specification.markdown
```diff
@@ -97,242 +170,298 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+
+Some examples of the encoding of integers of width 1 to 4. The x:es
+represent bits where the actual integer value would be stored.
+
+Width | Size | Representation
+------|------|----------------------------------------
+1     | 2^7  | 1xxx xxxx
+2     | 2^14 | 01xx xxxx xxxx xxxx
+3     | 2^21 | 001x xxxx xxxx xxxx xxxx xxxx
+4     | 2^28 | 0001 xxxx xxxx xxxx xxxx xxxx xxxx xxxx
```
robUx4: You may set the alignment as a set of 0

specification.markdown
```diff
@@ -97,242 +170,298 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+
+## Elements semantic
+
+### Element Template
+
+Element Name:
+
+    Level:
```
robUx4: Explain that a + means it's nested. It may also be another line:
`Nested: Yes/No`

specification.markdown
dericed: In the Element Template, we could add a value called "Element Status", then for each signature value, we can set it to "Deprecated since EBML version 1". Any better idea?

specification.markdown
dericed: fixed in https://github.com/MediaArea/ebml-specification/commit/db1042fb059c14a34578c025462bc90c973e7a95

dericed:
@mbunkus I'm going to read/respond/commit to the rest of the comments from @robUx4 and you, then will skim the commit list for squash opportunities, feel free to point them out :)

specification.markdown
dericed: [RFC 2223](http://tools.ietf.org/html/rfc2223#section-3) states:
    
     Each line must be limited to 72 characters followed by carriage return and line feed.

but even at 72 there still isn't a reason for a line break here, fixed in https://github.com/MediaArea/ebml-specification/commit/26b506c69a22e814ce9bbfa600a3ffb71113054c

specification.markdown
dericed: Add an issue for now, back to this later: https://github.com/Matroska-Org/ebml-specification/issues/7

specification.markdown
dericed: @mbunkus, are you ok if 'Class ID' changes to 'Class Name' here as well?

    The leading bits of the Class IDs are used to identify the
        length of the Element ID. The number of leading 0's + 1 is the
        length of the ID in octets. We will refer to the leading bits
        as the Length Descriptor.

specification.markdown
dericed: So to clarify, I should remove this commit from the PR? Or shorten it to remove the 'shortest encoding form' requirement?

dericed:
@robUx4 agreed about the mix of formatting and change, but not certain of the best strategy to split one PR into 2

robUx4:
@dericed I'm afraid you have to make 2 PRs. Whichever requires the less discussion should come first so we can move forward quickly. So that would be formatting. After that each *content* change can be discussed in individual PRs that can be merged independently. Especially if the changes don't occur in the same lines.

specification.markdown
robUx4: Sounds okay to me.

specification.markdown
robUx4: Yes, it should go in its own PR.

dericed:
@robUx4, np, I'd rather do it, right. I'll send start to send smaller sectional pull requests, first content, then formatting.

dericed:
Update: I've been squashing the commit history here, so it's down to 30 commits from 44 or so. This PR is not yet complete. A formatting-only PR here https://github.com/Matroska-Org/ebml-specification/pull/8 should be considered before this one.

robUx4:
Formatting PR merged

dericed:
few more commits merged into https://github.com/Matroska-Org/ebml-specification/pull/9

dericed:
two more pull requests based on the Element ID and Element Data Size are at https://github.com/Matroska-Org/ebml-specification/pull/10 and https://github.com/Matroska-Org/ebml-specification/pull/11

PR #9, opened 2015-05-10
labelling updates in semantic table

dericed:
as discussed at https://github.com/Matroska-Org/ebml-specification/pull/3


mbunkus:
Looks good to me. Thanks.

PR #11, opened 2015-05-12
add in draft definition of Element ID

dericed:
based on the rfc draft but using commas in numerical representation and correcting a reference to EBMLMaxIDLength

see prior conversation occurred on this at https://github.com/Matroska-Org/ebml-specification/pull/3


dericed:
As mentioned on IRC the patch contains some parts of the copywritten RFC Draft. I'll rewrite the core concepts, rebase, and re-open.

PR #13, opened 2015-05-27
Variable size integer

dericed:
This rewrites a definition on Variable Size Integer that is compatible with the license of the specification.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -40,6 +40,76 @@ to make up an EBML "document". Elements incorporate an Element ID, a
 descriptor for the size of the element, and the binary data itself.
 Futher, Elements can be nested or contain Elements of a lower "level".
 
+## Variable Size Integer
+
+The Element ID and Element Data Size are both encoded as a Variable
+Size Integer, developed according to a UTF-8 like system. The Variable
+Size Integer is composed of a VINT\_WIDTH, VINT\_MARKER, and VINT\_DATA,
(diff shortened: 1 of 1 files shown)


dericed:
@mbunkus I used VINT for most later references to Variable Size Integer and fixed the overlong line, then rebased and squashed. New commit is 623eb5c.

dericed:
I'd also like to verify that "Variable Size Integer" is a proper EBML term. The term doesn't exist in the current specification but seems to have been introduced in the RFC draft. Also I don't see any reference to the term in libebml. Is this the right term?

mbunkus:
I… am not sure. I definitely think it's the right one, but I don't remember ever having a real discussion about it. So unless @robUx4 objects I'd say: go for it.

dericed:
I added this line

    Each VINT component (VINT\_WIDTH, VINT\_MARKER, and +VINT\_DATA) MUST be expressed a big-endian unsigned integer.

then rebase and squashed. Note that although this statement seems well presumed I didn't see a pre-existing statement to support this from the RFC draft or elsewhere.

wm4:
Maybe it should be mentioned that "overlong" sequences are ok. (UTF-8 does not allow them, and for some reason this text says this encoding "UTF-8 like".)

mbunkus:
Looks good enough for me for now. Thanks.

PR #14, opened 2015-05-28
expand section on Element IDs

dericed:


dericed:
"MAY NOT" -> "MUST NOT" then rebased

dericed:
I re-synced my element-id-rewrite branch against Matroska-Org/master and force pushed, so that the PR can be pulled without conflict.

mbunkus:
Thanks.

PR #15, opened 2015-05-28
expand section on Element Data Size

dericed:


dericed:
- Note: the rules for unknown Data Size (that the master-element ends at the next element which is invalid at the sub-level of the master-element), presumes that the EBML Schema defines each Element and what levels it is valid at. I can add this info to the semantic section later.

- I want to clarify that empty EBML Elements are allowed. The spec lists 0 as the bottom range of valid range for data sizes, but I think there should be a specific statement on zero-sized elements. Are there any odd notes about zero-sized elements?

mbunkus:
Zero-sized elements are definitely allowed; it often happens in Matroska with master elements whose children are all set to their default value (and therefore omitted from the file). It's fine to state that such elements are indeed valid.

Note that I'm not sure if e.g. signed or unsigned integers with a data size of 0 octets are actually considered valid… libEBML always writes at least one octet for the data size, even if the data to encode is the value 0. Judging from the code libEBML should be able to read an unsigned integer with a data size of 0 octets; that would result in an unsigned integer of value 0.

dericed:
I added a paragraph on when vint-data is zero filled in the Element Data Size as referred to Elements with this case as "Empty Element".

dericed:
>Note that I'm not sure if e.g. signed or unsigned integers with a data size of 0 octets are actually considered valid… libEBML always writes at least one octet for the data size, even if the data to encode is the value 0. Judging from the code libEBML should be able to read an unsigned integer with a data size of 0 octets; that would result in an unsigned integer of value 0.

My opinion, but I would suggest that a 0-octet signed/unsigned integer would have a different meaning than where Element Data is zero-filled. The latter explicitly means zero whereas the former would be unknown (perhaps invalid depending on the EBML Schema).

This would be similar in XML to:

```
<expected_unsigned_integer></expected_unsigned_integer>
```
vs
```
<expected_unsigned_integer>0</expected_unsigned_integer>
```

dericed:
> As you've said yourself a clarification that the list of known elements is defined by the semantics layer above EBML would be beneficial here. A pure EBML-only parser would not be able to parse such an element properly; parsing is only possible in conjunction with the semantic information.

I would propose that the EBML Schema define which Elements MAY have an unknown size, but this is change from version 1. Maybe to put on the wishlist for version 2 of EBML. At any rate I added some language on the dependency on the schema for the handling of unknown sizes.

Note I intend to add a section on EBML Schema soon that defines the requirements for declaring one.

dericed:
changes made and rebased, please re-review

mbunkus:
Elements without VINT_DATA bytes: hmm, yes, in XML these are different values semantically. However, I'd argue that we cannot or should not introduce a concept of an undefined value now. All of EBML's (and Matroska's) elements have defined values; introducing something fundamentally different is bound to lead to problems.

I think we should either state that integer/floating point elements without VINT_DATA represent the value 0 or that such elements are invalid. Not sure what would be better; probably the »value 0« case.

mbunkus:
Just to be clear: certain EBML element types (strings, binaries, masters) are perfectly fine without VINT_DATA: empty strings, empty binaries, a master without children. I'm only talking about the scalar ones (signed integer, unsigned integer, float, date).

dericed:
Suggestions on wording to address empty VINT_DATA? :)

wm4:
So what's the call on VINT_DATA with length 0? I know my own parser (derived from mplayer's) considers it invalid.

mbunkus:
@wm4 Actually there's nothing ambiguous about it. Even [the current specs](https://github.com/Matroska-Org/ebml-specification/blob/master/specification.markdown) say that u/s integers need at least one byte of VINT_DATA, float needs 4 or 8.

Therefore: they're invalid.

specification.markdown
```diff
@@ -64,30 +64,67 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+An Element Data Size with all VINT\_DATA bits set to one is reserved as
+an indicator that the size of the Element is unknown. This rule allows
+for an Element to be written and read before the size of the Element is
+known; however unknown Element Data Size values SHOULD NOT be used
+unnecessarily. An Element with an unknown Element Data Size MUST be a
+Master-element in that it contains other EBML Elements as sub-elements.
+The end of the Master-element is determined by the beginning of the next
+element that is not a valid sub-element of the Master-element. The use
+of Elements of unknown size is dependent on the definition of the EBML
+Schema declared in DocType, because an Element of unknown size can not
```
robUx4: This is incorrect. You can handle an infinite element without knowing the "full" schema. Because EBML is forward compatible, new elements could be added to the file and older parsers should still be fine. It's not possible to handle infinite elements without know the base Schema. But that's the same for everything else in EBML.

specification.markdown
```diff
@@ -64,30 +64,67 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+3            | 0 to 2^21-2
+4            | 0 to 2^28-2
+5            | 0 to 2^35-2
+6            | 0 to 2^42-2
+7            | 0 to 2^49-2
+8            | 0 to 2^56-2
+
+If the length of Element Data equals 2^(n*7)-1 then the octet length of
+the Element Data Size MUST be at least n+1. This rule prevents an
+Element Data Size from being expressed as a reserved value. For example,
```
robUx4: There's only one reserved value and that's the infinite size. It would be better to mention it as it is to avoid confusion and complexity.

specification.markdown
wm4: We've already discussed this on the mailing list, and came to the conclusion that it's better to restrict use of elements with unknown length.

> It's not possible to handle infinite elements without know the base Schema. But that's the same for everything else in EBML.

Without unknown length elements, you can skip unknown subelements, making it future extensible, which is pretty much the whole point of EBML.

robUx4: I meant without a Schema at all you cannot do much with an EBML file.

robUx4: Also I looked quickly in the mailing list for infinite/unknown discussions. I do not see where we reached a conclusion on limiting it. I repeatedly said there's no need for this restriction.

dericed:
I expanded on the validity of Empty Elements on how it is conditional per Data Type. To summarize Empty Elements are valid with String, UTF-8, Master-element, and Binary and invalid with Signed Integer, Unsigned Integer, Float, and Date. Will rebase and push shortly.

specification.markdown
dericed: @robUx4 I changed
> An Element Data Size with all VINT\_DATA bits set to one is reserved as
> an indicator that the size of the Element is unknown. 

to

> An Element Data Size with all VINT\_DATA bits set to one is reserved as
an indicator that the size of the Element is unknown. The only reserved
value for the VINT\_DATA of Element Data Size is all bits set to one.

dericed: 1d579b615edcb5520bbaa9a5ecb466f88d947da4

robUx4: ok

specification.markdown
```diff
@@ -64,30 +64,78 @@ followed by the Data Size, and then the non-interpreted Binary itself:
...
+This rule allows for an Element to be written and read before the size
+of the Element is known; however unknown Element Data Size values SHOULD
+NOT be used unnecessarily. An Element with an unknown Element Data Size
+MUST be a Master-element in that it contains other EBML Elements as
+sub-elements. The end of the Master-element is determined by the
+beginning of the next element that is not a valid sub-element of the
+Master-element. The use of Elements of unknown size is dependent on the
+definition of the EBML Schema declared in DocType, because an Element of
+unknown size can not be parsed without a complete list of all possible
+sub-elements.
```
robUx4: Again, it can be parsed with a partial list of all possible sub-elements and even a partial list of upper elements. Unknown EBML elements will just be skipped until something known (immedialy lower or higher) is found.
Without this EBML files would not be backward compatible.

mbunkus: Please read my email about this on the devel ML where I present a case in which such partial knowledge will not be enough.

dericed: Sorry to take the conversation to two places. For those that happen upon this discussion here's a link to the post that @mbunkus mentioned. http://lists.matroska.org/pipermail/matroska-devel/2015-June/004703.html

dericed: can we proceed if I remove this line?
> The use of Elements of unknown size is dependent on the
definition of the EBML Schema declared in DocType, because an Element of
unknown size can not be parsed without a complete list of all possible
sub-elements.

mbunkus: I'd be fine with that.

mbunkus: Meaning: I think you addressed everything else I had pointed out, and that sentence was the only controversial part.

dericed:
I updated the PR to remove the statement about parsing unknown size being dependent on schema knowledge. At this point I think this clears all issues brought up in the discussion. Ready to pull in my view.

mbunkus:
Unless @robUx4 objects I'll merge this on Friday (I'll be out of town until then anyway).

mbunkus:
As there have been no objections I'm merging it now.

PR #16, opened 2015-05-28
remove matroska reference from EBML spec

dericed:
This PR removes references of matroska from the EBML spec. The EBML specification is not intended to be dependent on the Matroska specification. The restriction mentioned here that Matroska adds to the EBML specification (such as max Element Data Size lengths) should be moved from here to the Matroska spec.


mbunkus:
Thanks.

PR #17, opened 2015-06-27
extend definition of EBML Data Types

dericed:


specification.markdown
```diff
@@ -14,24 +14,6 @@ Just like XML, the specific "tags" (IDs in EBML parlance) used in an
 EBML implementation are arbitrary. However, the semantic of EBML
 outlines general data types and ID's.
 
-The known basic types are:
```
dericed: Note that moved the location of the section on data types to later in the specification.

specification.markdown
```diff
@@ -228,6 +210,95 @@ VINT\_WIDTH | VINT\_MARKER | VINT\_DATA     | Element Data Size Status
...
+## EBML Element Types
+
+Each defined EBML Element MUST have a declared Element Type. The Element Type defines a concept for storing data that may be constrained by length, endianness, and purpose.
+
+Element Data Type:   Signed Integer
+
+    Endianness:     Big-endian
+    Length:         A Signed Integer Element MUST at least 1 octet in
+                    length and no more than 8 octets in length.
+    Definition:     A Signed Integer stores an integer (meaning that it
```
dericed: Is there a good external definition of Signed Integer, Unsigned Integer, and Float to cite instead of defining them locally here.

specification.markdown
```diff
@@ -228,6 +210,95 @@ VINT\_WIDTH | VINT\_MARKER | VINT\_DATA     | Element Data Size Status
...
         0xFE and 5 can be coded 0x000000000005 or 0x0005 or 0x05.
 
+## EBML Element Types
+
+Each defined EBML Element MUST have a declared Element Type. The Element Type defines a concept for storing data that may be constrained by length, endianness, and purpose.
+
+Element Data Type:   Signed Integer
+
+    Endianness:     Big-endian
+    Length:         A Signed Integer Element MUST at least 1 octet in
```
dericed: Note: this is some controversy on whether Signed or Unsigned Integers are 0 length are valid.

specification.markdown
```diff
@@ -228,6 +210,95 @@ VINT\_WIDTH | VINT\_MARKER | VINT\_DATA     | Element Data Size Status
...
+                    defined in IEEE 754.
+
+Element Data Type:   String
+
+    Endianness:     None
+    Length:         A String Element may use any length (included zero)
+                    up to the maximum Element Data Size value permitted.
+    Definition:     A String Element may either be empty (zero-length)
+                    or contain Printable ASCII characters in the range
+                    of 0x20 to 0x7E. Octets with all bits set to zero
```
dericed: Note: some controversy about why zero-padding is allowed, but all prior docs seem consistent about allowing it.

specification.markdown
```diff
@@ -228,6 +210,95 @@ VINT\_WIDTH | VINT\_MARKER | VINT\_DATA     | Element Data Size Status
...
+
+Element Data Type:   Date
+
+    Endianness:     None
+    Length:         A Date Element MUST be 8 octets in length.
+    Definition:     The Date Element MUST contain a Signed Integer that
+                    expresses a point in time referenced in nanoseconds
+                    from the precise beginning of the third millennium
+                    of the Gregorian Calendar in Coordinated Universal
+                    Time (also known as 2001-01-01T00:00:00.000000000
```
dericed: I'm curious why is the Date Element centered at 1/1/2001 rather than the Unix epoch? Is the reasoning worthy of a note in the spec.

mbunkus: The Unix epoch is due to an API peculiarity on one operating system. I guess we didn't like specifications to be based on specific APIs, and in that case the beginning of the current century (which is what 2001-01-01 is) seemed to be as good a date to base upon than any other.

specification.markdown
wm4: Two's complement is an ubiquitous and well-known concept enough, I guess.

For floats, you could cite the exact IEEE standard.

wm4:
I'm surprised that you declared 0-length integers elements invalid, while you declared zero-padded strings valid. Maybe it's better to tell the truth, and write something like "0-length integers can occur in files produced by legacy software, but should not be produced by conforming software", same for strings.

specification.markdown
wm4: I guess it could be defined in terms of a single offset to the Unix epoch.

specification.markdown
dericed: Thanks, I'll add a reference to IEEE 754.

dericed:
@wm4 You voted no to 0 length integers here: http://lists.matroska.org/pipermail/matroska-devel/2015-June/004709.html. See also the discussion on https://github.com/Matroska-Org/ebml-specification/pull/17 particularly: https://github.com/Matroska-Org/ebml-specification/pull/15#issuecomment-107623446. I defer to others on this decision but want to be clear that allowing zero length integers is a change from the current official specifications.

specification.markdown
robUx4: EPOCH 32 bits was not precise enough. So since we needed something else, we used a different reference point to avoid confusion and easily point out errors. The beginning of the millennium seemed like a fancy date.

specification.markdown
robUx4: Old specs clearly mentioned 1 to 8 octets. So while some implementations may allow it, some may not. It only matters for elements that have a default value but are not mandatory (rare). The `libebml` code allows size 0 for reading and writing.
In general the code was designed to not write the value if the value is the default value, regardless of the type. That means size 0 is always allowed for elements with a default value. That's from the original of saving space in the output file for stuff we don't need to store.
Since String elements were already allowed to be of 0-length, I'd say there shouldn't be much code that will break with non default 0-length elements. So I vote in favor of allowing it globally. And explaining what it means and when it can be used (default value otherwise not allowed).

wm4: > That means size 0 is always allowed for elements with a default value. 

Are you saying that size 0 elements actually imply the default value, instead of the value 0?

robUx4: Mh, looking at the code of libebml, when reading an integer the default value is set to 0. But I can't find in the logic if the element data is actually read when the size is 0 or not. It's possible the value is not actually read and when picking the value for the element, the default value is used.

@mbunkus can you confirm which way you use it ? In https://github.com/Matroska-Org/libebml/blob/master/src/EbmlMaster.cpp#L423 the `MaxSizeToRead` can be 0 and the element is Found, but not added to the `ElementList` which means it's lost. That seems like a bug to me.

I looked in `libebml2` and the there when the value of an integer is 0, the value 0 is used.

Also in the library there are checks in place so that an element which wasn't set and is not mandatory will crash. So an integer with a size of 0 will never be written by `libebml`, unless it's mandatory.

mbunkus: If the data size is 0 then the `SetValueIsSet()` is called, too, see e.g. https://github.com/Matroska-Org/libebml/blob/master/src/EbmlUInteger.cpp#L142

This is exactly what I would expect. If an unsigned integer (or signed) is present but has a data size of 0 it is handled the exact same way as if the data size was > 0 and the data portion equals 0. libebml will therefore treat (un)signed integers with data size 0 as being present. Matroska's default values do not come into play in such a case.

This is also how I've always understood the specs. A default value is used if and only if the element is not present in the file at all. The element's data size does not influence that decision.

I'm fine with allowing (un)signed integers with a data size of 0. It takes away some ambiguity; therefore it makes parsers a bit easier to write. Additionally it saves a byte in very rare cases (though the latter is more a nice side effect from my point of view and not a rationale for allowing them).

robUx4: OK, so a size of 0 should be allowed and the interpretation depends on the type specified in the semantic. That is a value of 0 for signed/unsigned integers, floats, dates and an empty UTF-8/string. For a master or binary it doesn't have much sense.

mbunkus: Binary and master: not quite correct. For a binary a length of 0 may make sense in the theoretical case in which there's a default value for a binary element.

More important is the master case. This can actually happen in Matroska! Let's assume you have an audio track with a sampling frequency of 8000 Hz and one channel. In such a case the file must contain a TrackAudio master element, but that master may not contain any element as 8000 is the default value for SamplingFrequency and 1 is the default value for Channels.

This is exactly what happens if mkvmerge writes such a track as it sets libebml to skip elements set to their default values. Note that leaving out the TrackAudio element would NOT be the same thing semantically as elements with default values only apply if their parent element is present in the file.

robUx4: If we assume the default value for binary, it would make more sense to do the same for other types too, which is not the case right now. It would be better to have a consistent behavior. Maybe decide the value differently if the element has a default value or not ?

About Master, it's true that forcing the presence of an element with all default values is different than not having the element at all.

dericed: Here are some notes to ensure that I haveve understood this conversation correctly.

- Consensus now is that 0 length data will be allowed for unsigned and signed integers.

- Shall we then also allow 0 for floats with a meaning of 0?

- Zero length dates seem a bit ridiculous. I suggest not allowing a zero length date.

- A zero length element shall indicate a meaning declared in the 'data type' section of the ebml spec. A zero length element is not intended to be interpreted to the default value of that element (the default is used for elements that are missing AND mandatory).

- Whether or not an element uses a default is defined in the semantics rather than here in the definitions of data types.

mbunkus: Sounds about right to me. For the sake of uniformity I would vote for allowing zero-length date fields. Those equal 0 which in turn means the beginning of the century. Having less exceptions is better, IMHO. Just say that all data types may have a data size of 0 octets, and 0 octets means that all numeric types (signed/unsigned integers, floats) equal 0, the date type equals 0 which is semantically the beginning of the century, the string and binary types equal empty strings/binary elements and for masters it means that there are no child elements.

This is how most programming languages work, too, if they initialize objects.

Less exceptions = less code to test, less things to misunderstand.

wm4: >Less exceptions = less code to test, less things to misunderstand.

To be fair, 0-sized elements that are integers/floats/dates are extremely rare (rare enough that your code will handle at least 99.9% of all mkv files without breaking), so allowing this actually adds more need to test obscure corner cases.

mbunkus: I'm not really convinced of that. One if not the most often used way of reading an element goes something like this:

    data_size = read_data_size
    value = 0
    while data_size > 0:
        value += read_byte # (adjust for different data types)
        --data_size

So you always initialize the value anyway, and not only if the data size is != 0.

One thing I sincerely detest about C/C++ is the multitude of cases that cause undefined behavior (like signed integer overflow). Undefined behavior means that the compiler can assume anything, and that leads to so many subtle bugs on occurring in certain cases. I would really like to avoid such cases in EBML if possible. Saying that a data size of 0 is not allowed – what does that mean? That the element itself is invalid and must be ignored? That it can be treated as if it wheren't present? That the parser can return the element containing a random value (e.g. by not initializing its value field unless the data size is > 0)?

I'd like this case to be defined clearly so that such questions cannot even arise. Make it easy for implementers.

Also about the test cases. You can indeed argue that handling `data_size == 0` at the parser level is one more case to test. However, if you forbid this situation the burden of testing moves a level higher. If your parser library doesn't handle such elements then the application above it will probably need some tests for this corner case. If such elements represent undefined behavior then the parser library might simply return an element whose data size is 0 but whose value field is undefined/uninitialized/contains garbarge. An application would have to make sure it can cope with such elements gracefully.

This puts the burden of testing upon many more shoulders (as there's 1:n relationship between the parser library and applications using it) than if it were the parser library's job.

Undefined behavior is bad. In C/C++ there are so many cases of UB only because those languages aim to provide maximum efficiency and have a »don't pay for what you don't use« policy. This is not EBML's concern, though; reliable, extensible and flexible storage is. Those goals conflict with undefined behavior.

Yes, I do feel strongly about that.

wm4: Well, the codepath rejecting 0-sized elements wouldn't be different from the codepath rejecting otherwise invalid data (like overlong elements). I can't quite follow why we suddenly need to explicitly allow this corner-case for this reason. But let's stop this discussion.

dericed: This particular corner case issue is prominent because of inconsistencies between Martin Nilsson's RFC Draft of Matroska and the official specs. The draft allowed many zero-signed EBML data types that the official EBML/MKV specs didn't. From what I'm reading there still seems to be consensus for allowing zero-sized elements with integers and floats.

robUx4: I agree 0-length elements should be allowed.

And avoiding special cases for dates, even if it's never used. In libebml that would require adding code because when reading an element we read he ID, the length and we verify it's legit (otherwise the data are likely bogus and we don't use them). So we'd have to check if a date element has a size of 0. And in the end, why do we want to reject it ? (a date in libebml is an EBML signed integer on which we can read/write a date, I'd assume most implementations would do the same).

dericed: I hope I'm interpreting the discussion above correctly. I just committed https://github.com/MediaArea/ebml-specification/commit/fa5dc611122262431d2d29ed1d153627a0c905ab which adds descriptions of zero-length use of integers, floats, and dates. Should I say that these are allowed but not recommended?

dericed:
At this point, I think the new commits of the PR respond to the discussion above. Please re-review.

mbunkus:
I only found that one missing verb. Otherwise it seems OK to me now.

PR #22, opened 2015-08-31
set docType default to matroska

dericed:
This patch aligns the EBML spec and matroska spec. Some discussion on this issue is here: http://lists.matroska.org/pipermail/matroska-devel/2015-July/004734.html with @robUx4 opting to set DocType default to matroska to match the matroska spec. Note that with this change, all EBML Header elements would have defaults, meaning that a valid matroska file could simply start with an empty EBML Master-element, such as 0x1A45DFA380 and an EBML parser would presume all default values.

Update: this forum seems to imply that opposite, that EBML's DocType with no defaults is correct and Matroska's spec with a 'matroska' default is wrong. https://bugzilla.mozilla.org/show_bug.cgi?id=778436#c12.

I suggest some discussion and/or review on this PR.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -395,7 +395,7 @@ Element Name:   DocType
     Mandatory:      Yes
     Multiple:       No
     Range:          -
-    Default:        -
+    Default:        matroska
     Element Type:   String
     Description:    A string that describes the type of document that
                     follows this EBML header.


mbunkus:
I rather agree with the WebM guys. The Matroska spec cannot (re-)assign default values to EBML elements, only to Matroska elements. Therefore I'd rather change the Matroska specs for v4.

robUx4:
There's a confusion here. Although EBML has a website and specs. The only document that had been maintained was the "Matroska specs" which are actually Matroska and EBML specs. The layout make it look like the Matroska specs are defining values in EBML, but it's just for easy layout creation and alignment. It really says what the EBML should do.

So because in most cases devs assumed that no DocType in EBML header meant it's a Matroska file, we can't remove the default anymore. These files would have no official DocType at all. There were also cases some players would refuse to play a file because it didn't have any DocType.

So I'mokay with this patch, unless @mbunkus disagrees.

mbunkus:
I'm fine with it, too.

PR #23, opened 2015-08-31
allow void element at zero-level

dericed:
This resolves a conflict between the EBML and MKV specs. See discussion at http://lists.matroska.org/pipermail/matroska-devel/2015-July/004734.html.


mbunkus:
Thanks.

robUx4:
LGTM.

PR #25, opened 2015-10-16
EBMLMaxIDLength clarifications

dericed:
clarifying the range of possible valid values


dericed:
Just a note about the change here. A removed statement said "The maximum length of the IDs you'll find in this file." I changed the term 'file' to 'EBML Document' since from my understanding a file could include a stream of more than one EBML Document and thus more than one EBML Header.

So a file could be:
EBML (max id length = 8)
HUUUUUUGE_ID
EBML (max id length = 4)
SEGMENT

mbunkus:
Looks good to me, thanks.

PR #27, opened 2015-10-16
add definitions of EBML Document, Header, and Stream

dericed:
needs review

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -3,7 +3,7 @@
 ## About
 
 This repository contains the formal and official specification for the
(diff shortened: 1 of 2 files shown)


dericed:
This PR starts to tackle some of the things about EBML that are presumed but unwritten. Part of this is inspired by a conversation between myself and @robUx4 at VDD regarding defining EBML Document and EBML Stream so that requirements or constraints could be associated with these concepts individually when appropriate. Please review.

dericed:
scrapping this PR, will resubmit soon

PR #28, opened 2015-10-16
define EBML Document, Stream, Header

dericed:


dericed:
This is a resubmit of https://github.com/Matroska-Org/ebml-specification/pull/27 which I closed because of bad PR errors.

This PR starts to tackle some of the things about EBML that are presumed but unwritten. Part of this is inspired by a conversation between myself and @robUx4 at VDD regarding defining EBML Document and EBML Stream so that requirements or constraints could be associated with these concepts individually when appropriate. Please review.

mbunkus:
Looks good to me, thanks. Sorry for not reading all diffs fully before posting that superfluous line-wrap-request.

PR #29, opened 2015-10-16
Fix to binary defintiion

dericed:


mbunkus:
Looks good to me, thanks.

PR #30, opened 2015-10-16
fix inconsistency about references to Empty Elements

dericed:


mbunkus:
Looks good to me, thanks.

PR #34, opened 2015-10-20
Ebml body

dericed:


specification.markdown
```diff
@@ -372,7 +387,8 @@ Element Name:   EBMLVersion
     Range:          -
     Default:        1
     Element Type:   Unsigned Integer
-    Description:    The version of EBML parser used to create the file.
+    Description:    The version of EBML parser used to create the EBML
+                    Document.
```
dericed: Note that these changes from file to EBML Document change the meaning of the spec, but hopefully make it more clear. If a file is an EBML Stream and contains many EBML Documents and thus many EBML Headers, the prior language implied that the EBML Header elements defined the whole file. In the case where the file is an EBML Stream with multiple EBML Headers this led to a contradictory interpretation. I think the references to the EBML Header refering to the EBML Document are more correct than reference the EBML Header to the file.

specification.markdown
```diff
@@ -328,6 +329,20 @@ with a length greater than 4 octets. All EBML Elements within the EBML
...
+the EBML Body. The end of the EBML Body, as well as the end of the EBML
+Document that it is contained within, may be considered as whichever
+comes first: the beginning of a new level 0 EBML Header or the end of
+the file. The EBML Body MAY only consist of EBML Elements and MUST NOT
+contain any data that is not part of an EBML Element. Although the EBML
+specification itself defines precisely what EBML Elements are to be
+used within the EBML Header, the EBML specification does not name or
+define what EBML Elements are to be used within the EBML Body. The
+definition of what EBML Elements are to be used within the EBML Body is
+defined by an EBML Schema.
```
dericed: EBML Schema is to be defined. I think to define it I'll start by extending the section called `EBML Template` to define the element descriptors more specifically. However to define what a proper expression of an `EBML Schema` is goes back to some open listserv discussions about specdata.xml and how to express a machine-readable `EBML Schema`.

specification.markdown
mbunkus: I agree with this.

dericed:
response to @mbunkus's comment on a vague 'it': https://github.com/MediaArea/ebml-specification/commit/0ec80d2dafc00311acb8d05308b0a2c5e5260aba

PR #37, opened 2015-10-26
remove cosmetic line breaks that force a 72 char line limit

dericed:


robUx4:
Make sure paragraphs don't get much bigger than they are otherwise we'll lose in changes visibility and readability of the specs.
Merging.

robUx4:
Mh, it seems the Github formatting of tabbed (code) text doesn't handle well long lines.

dericed:
i agree. I've been thinking that formatting may need to be reconsidered. I'll open a ticket for it.

PR #40, opened 2015-12-05
Ebml schema

dericed:
This PR touches on many conversations within matroska-devel, github, and (soon) cellar. Please review the section on mandatory/default carefully.

Also note that the EBML Schema section currently defined the current instance of the Matroska which is a machine-readable XML definition of the Matroska EBML DocType as seen here https://github.com/Matroska-Org/foundation-source/blob/3414f292bb5b1ac5ee2fd99976b7274fd81e48ee/spectool/specdata.xml. I hope to advance specdata.xml more closely to practices used in XML Schemas, but syncing the existing documentation with the existing specdata.xml seems like a good place to start.


specification.markdown
```diff
@@ -173,21 +173,55 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
 
-    Level:
-    EBML ID:    []
-    Mandatory:  [Mandatory]
-    Multiple:   [For level 0 Elements, this attribute defines whether or not the Element may be used multiple times within an EBML Document. For Elements at Level 1 or greater, this attribute defines whether or not the Element may be used multiple times within its parent Element.]
-    Range:
-    Default:    [Default value if the element is not found]
-    Element Type:
-    Description:
+As an XML Document the EBML Schema MUST use `<table>` as the top level element. The `<table>` element may contain `<element>` sub-elements. Each `<element>` node defines one EBML Element through the use of several attributes. Each attribute of the `<element>` node of the EBML Schema is defined here along with a note to say if the attribute is mandatory of not. EBML Schemas many contain additional attributes to extend the semantics but MUST not conflict is the definitions of the `<element>` attributes defined here.
```
mjbshaw: "mandatory **of** not" → "mandatory **or** not"

specification.markdown
```diff
@@ -173,21 +173,55 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+
+#### EBML Schema Element Attributes
+
+Within an EBML Schema the `<element>` uses the following attributes to define an EBML Element.
+
+| attribute name | required | definition |
+|----------------|----------|------------|
+| name           | Yes      | The official human-readable name of the EBML Element. The value of the name MUST be in the form of an NCName as defined by the [XML Schema specification](http://www.w3.org/TR/1999/REC-xml-names-19990114/#ns-decl). |
+| level          | Yes      | The level notes at what hierarchical depth the EBML Element may occur within an EBML Document. The initial EBML Element of an EBML Document is at level 0 and the Elements that it may contain are at level 1. The level MUST be expressed as an integer; however, the integer may be followed by a '+' symbol to indicate that the EBML Element is allow valid at any higher level.  |
```
mjbshaw: "is allow valid" → ("is allowed" || "is valid")

specification.markdown
```diff
@@ -173,21 +173,55 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+| global         | No       | A boolean to express if an EBML Element MUST occur at its defined level be may occur within any Parent EBML Element. |
```
mjbshaw: "level **be** may" → "level **or** may"

specification.markdown
```diff
@@ -173,21 +173,55 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+| id             | Yes      | The Element ID expressed in hexadecimal notation prefixed by a '0x'. |
+| mandatory      | No       | A boolean to express if the EBML Element MUST occur if the Parent EBML Element is used. If the mandatory attribute is not expressed for that Element then that element is to be considered not mandatory. |
+| multiple       | No       | A boolean to express if the EBML Element may occur within its Parent EBML Element more than once. If the multiple attribute is false or the  multiple attribute is not used to define the Element then that EBML Element MUST not occur more than once with that Element's Parent EBML Element. |
+| range          | No       | For Elements which are of numerical types (Unsigned Integer, Signed Integer, Float, and Date) a numerical range may be specified. If specified that the value of the EBML Element MUST be within the defined range inclusively. |
```
mjbshaw: What is the format for `range`? Is it `x-y` or `[x,y]` or `x..y` etc.? Is whitespace significant? Must the left-hand value be `<=` the right-hand value (i.e. is `2-1` equivalent to `1-2`, or is `2-1` malformed?)? How should ranges like "not 0" (which is currently used) be encoded?

Are integers encoded in decimal notation?

Are floats encoded in the same way as C (i.e. `3.14e0`)? If so, is [hex-encoding](https://gcc.gnu.org/onlinedocs/gcc/Hex-Floats.html) supported?

Are dates encoded as signed integers, or as ISO 8601 (just to pick a format)?

I think for the `range` attribute to be machine readable this needs to be nailed down pretty tightly.

specification.markdown
```diff
@@ -173,21 +173,55 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+The value of the `<element>` shall contain a description that of the meaning and use of the EBML Element.
+
+#### Note on the Use of mandatory and default attributes to define EBML Elements
+
+If an EBML Element has a default value declared by an EBML Schema and the EBML Element's value is equal to the declared default value then that Element is not required to be present within the EBML Document if its Parent EBML Element is present. In this case, the default value of the EBML Element may be assumed although the EBML Element is not present within its Parent EBML Element.
+
+If an EBML Element has no default value declared by an EBML Schema and its Parent EBML Element is present than the EBML Element must be present as well. If an EBML Element has a default value declared by an EBML Schema and its Parent EBML Element is present and the EBML Element's value is NOT equal to the declared default value then the EBML Element MUST be used.
+
+This table clarifies is an EBML Element which is defined as mandatory MUST be written, according to if the default value is declared, is the value of the EBML Element is equal to the declared default value, and if the Parent EBML Element is used.
```
mjbshaw: "clarifies **is** an" → "clarifies **if** an"

"**is** the value" → "**if** the value"

specification.markdown
```diff
@@ -173,21 +173,55 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+| mandatory      | No       | A boolean to express if the EBML Element MUST occur if the Parent EBML Element is used. If the mandatory attribute is not expressed for that Element then that element is to be considered not mandatory. |
+| multiple       | No       | A boolean to express if the EBML Element may occur within its Parent EBML Element more than once. If the multiple attribute is false or the  multiple attribute is not used to define the Element then that EBML Element MUST not occur more than once with that Element's Parent EBML Element. |
+| range          | No       | For Elements which are of numerical types (Unsigned Integer, Signed Integer, Float, and Date) a numerical range may be specified. If specified that the value of the EBML Element MUST be within the defined range inclusively. |
+| default        | No       | A default value may be provided. If an Element is mandatory but not written within its Parent EBML Element, then the reader of EBML Docuemnt MUST consider the defined default value of the Element. EBML Elements that are Master-elements MUST NOT declare a default value. |
```
mjbshaw: "Doc**uem**nt" → "Doc**ume**nt"

specification.markdown
```diff
@@ -173,21 +173,55 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+| type           | Yes      | As defined within the (##EBML-Element-Types) section, the type MUST be set to one of the following values: 'integer' (signed integer), 'uinteger' (unsigned integer), 'float', 'string', 'date', 'utf-8', 'master', or 'binary'. |
+
+The value of the `<element>` shall contain a description that of the meaning and use of the EBML Element.
+
+#### Note on the Use of mandatory and default attributes to define EBML Elements
```
mjbshaw: The wording in this section conflicts with the EBML Element Types section and [how things have been done in the past](http://lists.matroska.org/pipermail/matroska-devel/2015-October/004838.html).

Consider, for example, the `CueBlockNumber` MKV element. It is a non-mandatory element with a default value of `1`. If it's encoded in the byte stream as an element with a zero-octet length, then the EBML Element Types section makes it sound like the value should be 0 ("a zero-octet length represents an integer value of zero"). There is no mention anywhere in the document that an Empty Element should use its default value.

The current wording makes it sound like only mandatory elements may have a default value, which conflicts with what is currently being done in MKV/WebM.

specification.markdown
dericed: For reference here is the list of unique range expressions within the [current Matroska EBML Schema](https://github.com/Matroska-Org/foundation-source/blob/3414f292bb5b1ac5ee2fd99976b7274fd81e48ee/spectool/specdata.xml).

- &gt; 0
- &gt;0
- 0-1
- 1-254
- not 0

dericed: @mjbshaw, the EBML Schema concept of `range` is analogous to the XML Schema concept of `restriction`, so I propose aligning with the [specification of the `restriction` expression](http://www.w3schools.com/Xml/schema_facets.asp) to the extent feasible.

To document what currently exists (only the short list above), I suggest we could say:

- a prefix of the string `not ` negates the value of `range`
- the `range` must be expressed in of the these forms:
    - `x-y` where x and y are integers and y must be greater than x
    - `>x` where x is an integer, meaning that the value must be greater than `x`
    - `x` where x is an integer, meaning that the value must equal `x` (in the current schema this is only used with negation via `not `)
- white space may appear for readability but is not meaningful
- `range` are only defined for integer ranges
- `range` does not support date or float ranges

These rules serve to document what is currently there. Once that is in place, I propose we can proceed to change current practice towards the rules of the XML Schema `restriction`.

dericed: @mjbshaw, or others who may know, the [specdata.xml file](https://github.com/Matroska-Org/foundation-source/blob/3414f292bb5b1ac5ee2fd99976b7274fd81e48ee/spectool/specdata.xml) is the only precedent expression of an EBML Schema that I know of. Is there a machine readable expression of webm anywhere? Or any other EBML Document Type?

mbunkus: You should also state that in the range format `x-y` the values `x` and `y` are inclusive.

dericed: Add rules regarding the expression of `range` at https://github.com/MediaArea/ebml-specification/commit/4a9f56b0f08f03682dc1174cbfcd606c12691e4c.

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
 
 ## Elements semantic
 
-### Element Template
+### EBML Schema
 
-Element Name:
+An EBML Schema is an XML Document that defines the properties, arrangement, and usage of EBML Elements that compose a specific EBML Document Type. The relationship of an EBML Schema to an EBML Document may be considered analogous to the relationship of an XML Schema to an XML Document. An EBML Schema MUST be clearly associated with one or many EBML Document Types. An EBML Schema must be expressed as well-formed XML. An EBML Document Type is identified by a unique string stored within the EBML Header element called DocType; for example `matroska` or `webm`.
```
robUx4: Maybe add links to XML and XML Schema specifications in here ?
If this document is going to be the basis for the final RFC, shouldn't the links be all grouped at the end ?

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+
+#### EBML Schema Element Attributes
+
+Within an EBML Schema the `<element>` uses the following attributes to define an EBML Element.
+
+| attribute name | required | definition |
+|----------------|----------|------------|
+| name           | Yes      | The official human-readable name of the EBML Element. The value of the name MUST be in the form of an NCName as defined by the [XML Schema specification](http://www.w3.org/TR/1999/REC-xml-names-19990114/#ns-decl). |
+| level          | Yes      | The level notes at what hierarchical depth the EBML Element may occur within an EBML Document. The initial EBML Element of an EBML Document is at level 0 and the Elements that it may contain are at level 1. The level MUST be expressed as an integer; however, the integer may be followed by a '+' symbol to indicate that the EBML Element is valid at any higher level.  |
```
robUx4: Maybe the '+' should be moved to a different field 'nested'. That will make parsing easier.

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+| global         | No       | A boolean to express if an EBML Element MUST occur at its defined level or may occur within any Parent EBML Element. |
```
robUx4: When a field is not required, we must define the default value. In this case it would be 'false'.

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+| id             | Yes      | The Element ID expressed in hexadecimal notation prefixed by a '0x'. |
+| mandatory      | No       | A boolean to express if the EBML Element MUST occur if the Parent EBML Element is used. If the mandatory attribute is not expressed for that Element then that element is to be considered not mandatory. |
+| multiple       | No       | A boolean to express if the EBML Element may occur within its Parent EBML Element more than once. If the multiple attribute is false or the  multiple attribute is not used to define the Element then that EBML Element MUST not occur more than once with that Element's Parent EBML Element. |
+| range          | No       | For Elements which are of numerical types (Unsigned Integer, Signed Integer, Float, and Date) a numerical range may be specified. If specified that the value of the EBML Element MUST be within the defined range inclusively. See the [section of Expressions of range](#expression-of-range) for rules applied to expression of range values. |
```
robUx4: We might restrict the type names strictly. At least their casing so parsing is easier.

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+| default        | No       | A default value may be provided. If an Element is mandatory but not written within its Parent EBML Element, then the reader of EBML Document MUST consider the defined default value of the Element. EBML Elements that are Master-elements MUST NOT declare a default value. |
```
robUx4: What do you "consider" ? That sounds like it's optional.

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+| type           | Yes      | As defined within the [section on EBML Element Types](#ebml-element-types), the type MUST be set to one of the following values: 'integer' (signed integer), 'uinteger' (unsigned integer), 'float', 'string', 'date', 'utf-8', 'master', or 'binary'. |
```
robUx4: That may be extended for custom types, in that case such defined elements should be considered like binary when parsing.
So maybe custom types should be defined as 'binary' and use a different field to define what the type is exactly about.

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+
+The value of the `<element>` shall contain a description that of the meaning and use of the EBML Element.
+
+#### Expression of range
+
+The `range` attribute MUST only be used with EBML Elements that are either `signed integer` or `unsigned integer`. The `range` attribute does not support date or float EBML Elements. The `range` expression may contain whitespace for readability but whitespace within a `range` expression MUST NOT convey meaning. The expression of the `range` MUST adhere to one of the following forms:
```
robUx4: While I understand it's odd to retrict the dates stored in a format, it may be useful for floats. Restricting to 0.0 to 1.0 or -1.0 to 1.0 is common.

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+
+    - `x-y` where x and y are integers and `y` must be greater than `x`, meaning that the value must be greater than or equal to `x` and less than or equal to `y`.
```
robUx4: Since a range is allowed for `signed integer` it should not be separated by a `-` as it means a negative cannot be an upper value.

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+    - `>x` where `x` is an integer, meaning that the value MUST be greater than `x`.
+    - `x` where `x` is an integer, meaning that the value MUST be equal `x`.
+
+The `range` may use the prefix `not ` to indicate that the expressed range is negated.
+
+#### Note on the Use of mandatory and default attributes to define EBML Elements
+
+If an EBML Element has a default value declared by an EBML Schema and the EBML Element's value is equal to the declared default value then that Element is not required to be present within the EBML Document if its Parent EBML Element is present. In this case, the default value of the EBML Element may be assumed although the EBML Element is not present within its Parent EBML Element.
+
+If an EBML Element has no default value declared by an EBML Schema and its Parent EBML Element is present than the EBML Element must be present as well. If an EBML Element has a default value declared by an EBML Schema and its Parent EBML Element is present and the EBML Element's value is NOT equal to the declared default value then the EBML Element MUST be used.
```
robUx4: I supposed it should be *MUST*. But an element with no default value that is not mandatory doesn't have to be present. It's not clear whether this paragraph is for mandatory elements or any elements. If so maybe the title should be *Note on the Use of default attributes to define EBML Mandatory Elements*

specification.markdown
```diff
@@ -173,21 +173,65 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+
+This table clarifies if an EBML Element which is defined as mandatory MUST be written, according to if the default value is declared, if the value of the EBML Element is equal to the declared default value, and if the Parent EBML Element is used.
+
+| Is the default value declared? | Is the value equal to default? | Is the Parent Element used? | Then is storing the EBML Element required? |
```
robUx4: Did we rule out elements that are not mandatory but have their default value ? In which case they can be written with a data length of 0.

specification.markdown
mjbshaw: Agree. I have to ask, though, what's the point of `level`? If specdata.xml is changed to use nested elements, then `level` seems like redundant information.

A separate field like `nested` could use the names of elements as its value, instead of an integer (which would be kind of nice, as you could then have elements that are legal in some but not all others).

specification.markdown
mjbshaw: Using `-` is still parseable and doesn't lead to ambiguity (i.e. `-5--1` is still understandable), but it does look a little funny, so I'm not particularly opinionated whether `-` is used or something else.

specification.markdown
dericed: My idea was to document the current specdata.xml, then when done, move on to document a new EBML Schema so that the transition from one to the other is clear. `level` is needed in specdata.xml since there is no nesting. Obviously once nested we could drop `level`. Perhaps documenting specdata.xml isn't so helpful and we should jump directly to EBML Schema.

specification.markdown
mjbshaw: How does using the `default` attribute work for things like `DisplayWidth`/`DisplayHeight`/`BlockDuration`? They have a default value that depends on another element. How should that be encoded? Should whitespace be ignored (`OutputSamplingFrequency`'s default value has a space in it: `Sampling Frequency`)?

How are default values supposed to be encoded? Right now strings are encoded as `default="eng"`, which means that there is room for conflict/ambiguity if they have a default value that matches an element's name (i.e. `default="English"`: does it refer to the string `"English"` or the element named `English` (assuming such an element exists)).

Also, for the record, default values that depend on far away elements are a bit annoying/painful (`BlockDuration`, specifically). The default value for `BlockDuration` is not a sibling/parent/grandparent. It's a second-cousin, which makes designing a parser more difficult as it now needs access to an otherwise unrelated, far away element. I personally would like to see more restrictions on defaults that only permit them to depend on sibling/parent elements, but I realize that's an unrealistic wish. I just wanted to share a particular pain point that parser implementers will experience. (side note: `BlockDuration`'s default value still refers to the nonexistent `TrackDuration` element in `specdata.xml`)

specification.markdown
dericed: I added the links at https://github.com/MediaArea/ebml-specification/commit/e706a1db60b6163ca2c4e86818a0a941b9b6f09a.
I suggest re-formatting to RFC style formatting of references and links be done when it's closer to done. For now I added an issue for this at https://github.com/Matroska-Org/ebml-specification/issues/41

specification.markdown
dericed: fixed at https://github.com/MediaArea/ebml-specification/commit/0b16c5aa826355d6fd8cf45be7b5ba24a0bd0589

specification.markdown
dericed: @robUx4, better word that 'consider' for here? 'interpret'?

dericed: For reference here are the current `default` values used in specdata.xml:

```
0
1
1.0
1000000
4
50
8
8000.0
PixelHeight
PixelWidth
Sampling Frequency
TrackDuration
eng
matroska
und
```

dericed: @mjbshaw I presume that `Sampling Frequency` is a typo. I placed a PR here https://github.com/Matroska-Org/foundation-source/pull/7.

specification.markdown
dericed: Works for me. I suggest we (eventually) ok this pull request to document the current implementation of EBML Schema (specdata.xml) and then in a later pull request, we'll propose a new form of EBML Schema to moreso align with the methods already in place within XML Schema. In XML Schema this case that would mean expressing a restriction like:
```xml
<xs:restriction base="xs:integer">
    <xs:minInclusive value="-5"/>
    <xs:maxInclusive value="-1"/>
</xs:restriction>
```

specification.markdown
dericed: I left it applying to only the integers since that was the case in specdata.xml. I suggest we (eventually) ok this pull request to document the current implementation of EBML Schema (specdata.xml) and then in a later pull request, we'll propose a new form of EBML Schema to moreso align with the methods already in place within XML Schema. In XML Schema this case that would mean expressing a restriction like:
```xml
<xs:restriction base="xs:float">
    <xs:minInclusive value="-1.0"/>
    <xs:maxInclusive value="1.0"/>
</xs:restriction>
```

specification.markdown
dericed: I noted in an issue for now at https://github.com/Matroska-Org/ebml-specification/issues/42. Since no existing specs use custom Element Types, this discussion can be moved outside of this pull request.

specification.markdown
dericed: You're right, this wasn't clear enough. I updated the section to be more clear that it is discussing Mandatory EBML Elements here https://github.com/MediaArea/ebml-specification/commit/2606dc5ac432c02d837c4df18bb94cdb08593f59.

specification.markdown
dericed: I added a section on non-mandatory Empty Elements with default values here https://github.com/MediaArea/ebml-specification/commit/6154c58bef669cd2c434b0122d5aac2f88f4f679.

dericed:
I think all issues (-1) have been addressed and this pull request is ready for another review.

The remaining issue, noted at https://github.com/Matroska-Org/ebml-specification/pull/40#discussion_r46981092, is about clarifying if the default is a string or a reference to the value of another element. I think this is an important one but I don't have a proposal for this yet.

Unfortunately there is no precedent for the latter in the XML Schema specifcation. The XML Schema allows for default values for elements or attributes but the default values themselves are simply strings. Because of this I'd like to inquire if there is another way to solve this problem as extending the functionality of an EBML Schema beyond the equivalent functionality of an XML Schema would prevent the possibility of converting EBML->XML for verification in that more common environment.

A possibility (which does seem to have an equivalent in XML Schema) would to use something like xpath (maybe epath for us). so:
```
default="./PixelWidth"
default="./PixelHeight"
default="./SamplingFrequency"
```

Regarding `BlockDuration` defaulting to `TrackDuration`, this is confusing since there is no element called `TrackDuration`. I presume the default is actually referring to `DefaultDuration`, so the relationship is like:
```
Cluster
    BlockGroup
        BlockDuration
Track
    TrackEntry
        DefaultDuration
```
So an xpath like expression would be like:
```
default="../../Track/TrackEntry/DefaultDuration"
```

We'd still need a way to determine if the `default` represents a string or related Element.

mbunkus:
I kind of like the proposal with XPath, but it has other problems. However, I think that we should discuss that in a separate issue and merge this pull-request as-is for now.

If no one objects I'll do that tomorrow.

dericed:
Sounds like to me. I agree that xpath-like references add too much complexity to a single PR and should be handed separately.

mbunkus:
Done :)

specification.markdown
robUx4: I think the Schema should follow the specs document, not the other way around. It's easier to read/comment for most people.

PR #46, opened 2016-01-06
CRC-32 definition update

dericed:
related to discussion with the IETF CELLAR working group at https://mailarchive.ietf.org/arch/msg/cellar/NY3dXLIBJlVAKaauehwdt2nqBt4

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -336,7 +336,7 @@ Element Name:   CRC-32
     Range:          -
     Default:        -
     Element Type:   Binary
-    Description:    The CRC is computed on all the data of the Master element it's in. The CRC element should be the first in it's parent master for easier reading. All level 1 elements should include a CRC-32. The CRC in use is the IEEE CRC32 Little Endian.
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -336,7 +336,7 @@ Element Name:   CRC-32
     Range:          -
     Default:        -
     Element Type:   Binary
-    Description:    The CRC is computed on all the data from the last CRC element (or start of the upper level element), up to the CRC element, including other previous CRC elements. All level 1 elements SHOULD include a CRC-32.
+    Description:    The CRC-32 Element contains a 32 bit Cyclic Redundancy Check value of all the Element Data of the Parent Element as stored except for the CRC-32 Element itself. The CRC element SHOULD be the first in its Parent Element for easier reading. All Elements at Level 1 of an EBML Document SHOULD include a CRC-32 Element as a Child Element. The CRC in use is the IEEE-CRC-32 algorithm as used in the ISO 3309 standard and in section 8.1.1.6.2 of ITU-T recommendation V.42, with initial value of 0xFFFFFFFF. The CRC value MUST be computed on a little endian bitstream and MUST use little endian storage.
```
robUx4: IMO it should say `MUST be the first in its Parent`, not SHOULD. That will make processing easier. If you have tight live constraint, you may not use CRC at all if you can't afford to buffer.

dericed:
pull request updated to include a mandate for CRC-32 as the first element within its parent when used. https://mailarchive.ietf.org/arch/msg/cellar/LcVK2PxrHJ04tZNCiBYL8smHCgk

PR #47, opened 2016-01-10
add identical attribute to EBML Element Definitions

dericed:
as discussed https://mailarchive.ietf.org/arch/msg/cellar/bVwTRKL0FMfbxlibqPqFE4QHo84


mbunkus:
LGTM.

PR #49, opened 2016-01-10
explain non-element data in master-elements and suggestion for 4-octe…

dericed:
…t level 1 ids

as discussed in this cellar thread https://mailarchive.ietf.org/arch/search/?email_list=cellar&q=test4.mkv+and+EBML+Elements+with+Unknown+Size


dericed:
updated to `…then all contained Elements must be valid at level 3`

mbunkus:
Thanks.

PR #51, opened 2016-01-22
Ebml schema refactor

dericed:
This PR follows upon https://mailarchive.ietf.org/arch/msg/cellar/kvKt9FagzziL94tNfF0Itye83HA. Please review.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -155,7 +155,7 @@ An EBML Document is comprised of only two components, an EBML Header and an EBML
 
 ### EBML Header
 
-The EBML Header is a declaration that provides processing instructions and identification of the EBML Body. The EBML Header may be considered as analogous to an XML Declaration. A valid EBML Document must start with a valid EBML Header.
+The EBML Header is a declaration that provides processing instructions and identification of the EBML Body. The EBML Header may be considered as analogous to an XML Declaration. All EBML Documents MUST begin with a valid EBML Header.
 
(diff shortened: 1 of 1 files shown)


mbunkus:
I've reviewed the commits and left my notes as line notes.

dericed:
Unfortunately the squashing seems to have lost @mbunkus initial comments. Here's a summary of what I just did:
- in maxOccurs definition say it is the number of occurrences that MAY occur (was MUST before)
- switch lang tag in EBML Schema documentation to use RFC 5646
- remove accidental line breaks
- grammar suggestions on the arrangement of <element> nodes in an EBML Schema
- remove unneeded maxOccurs from the EBML Schema example

Should be ready for a re-review.

dericed:
Updated with minver and maxver attributes for EBML Schema. Updated to say that the EBML Schema doesn't document the EBML Header. Add a version attribute to EBML Schema. See also: https://mailarchive.ietf.org/arch/msg/cellar/yXQSR2dfT4ytUvc7JcD_eZzeb7w.

mbunkus:
I've added all the comments that I could think of for the push from Feb 9. Apart from those it looks good to me.

dericed:
Fixes according to the current comments here are in https://github.com/MediaArea/ebml-specification/commit/cd1c54b54417c0fc07746e651d8c0bf2e3bbdf40. Should I squash?

mbunkus:
Yes, please. I'll merge afterwards.

dericed:
done

mbunkus:
Thanks.

PR #54, opened 2016-02-13
draft security considerations section

dericed:


specification.markdown
```diff
@@ -12,6 +12,26 @@ EBML uses a simple approach of building Elements upon three pieces of data (tag,
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
+## Security Considerations
+
+EBML itself does not offer any kind of security. It has nothing to do with authentication, it dose not provide confidentiality, only marginally useful and effective data integrity options (CRC elements).
```
tophf: typo: `dose` -> `does`

specification.markdown
```diff
@@ -12,6 +12,26 @@ EBML uses a simple approach of building Elements upon three pieces of data (tag,
...
+
+EBML does not provide any kind of authorization.
+
+Even if the semantic layer offers any kind of encryption, EBML itself may leak information at either the sematic layer (as declared via the DocType element) or within the EBML structure (you can derive the presence of EBML elements even with an unknown semantic layer with a heuristic approach; not without errors, of course, but with a certain degree of confidence).
+
+Attacks on an EBML reader may include:
+- Invalid Element IDs that are longer than the limit stated in the EBMLMaxIDLength Element of the EBML Header.
+- Invalid Element IDs that not encoded in the shortest-possible way.
```
robUx4: that *is* not

specification.markdown
```diff
@@ -12,6 +12,26 @@ EBML uses a simple approach of building Elements upon three pieces of data (tag,
...
+- Invalid Element IDs comprised of reserved values.
+- Invalid Element Data Size values that are longer than the limit stated in the EBMLMaxSizeLength Element of the EBML Header.
```
robUx4: Longer of unsupported, like some float sizes that are not supported or simply don't exist.

dericed: Wrote this to cellar as well, but Element Data Size values are not floats, not sure how this applies.

mbunkus: What @robUx4 means is an element size that doesn't match the allowed ones for the `EbmlFloat` element type: either 4 or 8. For example, in Matroska the sequence `0x78 b5 81 …` would be invalid as it refers to the "output sampling frequency" element, a float, but its length is declared as 1 byte only.

I'd rather add this as a separate bullet point, something like

    - Invalid Element Data Size values that aren't allowed for the Element type (e.g. a length of 1 byte for an `EbmlFloat` element)

specification.markdown
tophf: `that are not encoded` or simply `not encoded`

dericed:
Thanks to @tophf, @robUx4, @mbunkus. All comments were added, then I squashed and force push the commits. Please review.

mbunkus:
Just that one nitpick. If you fix that I'll merge it. Thanks.

mbunkus:
Thanks.

PR #63, opened 2016-05-19
Fixed Typo in "Textual expression of Floats"

kit90:
Fixed a typo regarding the meaning of '-' character in float range.


mbunkus:
Thanks.

PR #64, opened 2016-05-22
ebml schemas should redundantly define void and crc-32

dericed:


specification.markdown
```diff
@@ -207,7 +207,7 @@ An EBML Schema MUST declare exactly one Element at Level 0 (referred to as the R
 
 Elements defined to only occur at Level 1 are known as Top-Level Elements.
 
-The EBML Schema does not itself document the EBML Header, but documents all data of the EBML Document that follows the EBML Header. The EBML Header itself is documented by this specification in the [EBML Header Elements](#ebml-header-elements) section.
+The EBML Schema does not itself document the EBML Header, but documents all data of the EBML Document that follows the EBML Header. The EBML Header itself is documented by this specification in the [EBML Header Elements](#ebml-header-elements) section. The EBML Schema also does not document Global Elements defined by the EBML Specification (namely VOID and CRC-32).
```
mbunkus: VOID is not an abbreviation and should therefore not be uppercase but merely capitalized like in other places.

dericed:
@mbunkus updated, please re-review

mbunkus:
Thanks.

PR #68, opened 2016-06-26
add type attribute to <documentation> element in ebml schema

dericed:


specification.markdown
```diff
@@ -235,7 +235,7 @@ Within an EBML Schema the `<element>` uses the following attributes to define an
...
 
-The `<element>` nodes shall contain a description of the meaning and use of the EBML Element stored within one or many `<documentation>` sub-elements. The `<documentation>` sub-element may use a `lang` attribute which may be set to the RFC 5646 value of the language of the element's documentation.
+The `<element>` nodes shall contain a description of the meaning and use of the EBML Element stored within one or many `<documentation>` sub-elements. The `<documentation>` sub-element may use a `lang` attribute which may be set to the RFC 5646 value of the language of the element's documentation. The `<documentation>` sub-element may use a `type` attribute to distinguish the meaning of the documentation. Recommended values for the `<documentation>` sub-element's `type` attribute include: definition, rationale, usage notes, and references.
```
mbunkus: I'd prefer to have the four possible values for the `type` attribute to be placed in `code tags`, too, as that's supposed to be an enumeration of fixed strings.

dericed: I added `code tags` and force pushed.

mbunkus:
Thanks.

PR #67, opened 2016-06-26
define recursive attribute and clarify level attribute

dericed:

PR #69, opened 2016-06-29
rewrite for clarity on level use with global and recursive elements

dericed:
in response to this comment from an already merged pr https://github.com/Matroska-Org/ebml-specification/pull/67#discussion_r68805221


mbunkus:
Thanks.

PR #73, opened 2016-07-07
Rfc fixes

dericed:


mbunkus:
Hmm, when you change from writing an explicit number of bytes to a hexadecimal number then the document must also state that the Element IDs are stored/read in Big Endian order. The spec already states Big Endian for the content of various Element types (e.g. the Signed/Unsigned Integer ones), but not for the IDs.

dericed:
i added a note about big endian element ids

PR #76, opened 2016-07-11
add size attribute

dericed:
Potentially the size for docType could be set to >0 to prevent an Empty Element.


robUx4:
Looking fine but other elements that are not fille have a `-` as a value. Should Size be set for all elements or those elements could do without the unused `-` value ?

dericed:
Lately I've been feeling like removing the `-` values from the definitions. For instance, as the language of the specification has gotten a lot stricter, it starts to look like
```
type: string
default: -
```

Implies that the default is a literal `-`.

robUx4:
I think it would make things more readable. Since it's not a table, all cells don't need to be filled.

dericed:
This is related to a discrepancy between EBML and MKV. The [Matroska EBML Schema](https://github.com/Matroska-Org/matroska-specification/blob/gh-pages/ebml_matroska.xml) documents the Root Element of Matroska (Segment), but not the EBML Header itself. I think this is consistent with the EBML Schema/XML Schema analogy as XML Schemas do not document the XML Declaration (which is the equivalent of the EBML Header). Still presently we don't have a machine-readable structure for the EBML Header in either repository.

Possibly we could say that all EBML Schema's just include the EBML Header definition with the EBML Schema of each docType. Though this seems a bit redundant.

dericed:
Proposals for size contrains is welcome. I think there are many inferred limits that we can now better clarify.

robUx4:
`SegmentUID`, `PrevUID`, `NextUID`, `SegmentFamily` and all the elements referencing these values are 16 octets long.
`TrackType` is mentioned as 8 bits although there's no particular reason for that.
`ColourSpace` is 4 octets long as it's a value from AVI.

PR #78, opened 2016-07-29
Fixed typos and made small changes on word level

marionj1:
Thanks for the quick response. I was asked by Dave to try to include some changes that Tim T. proposed on the cellar mailing list. The reason for deleting the article before "section" was that in the document https://datatracker.ietf.org/doc/draft-lhomme-cellar-ebml/ the references read differently. For example "See the section of expression of range..." reads "See the Section 1.11.1.4"... I thought that if I apply the changes that Tim proposed there won't be troubles if there is a different display of the references like https://datatracker.ietf.org/doc/draft-lhomme-cellar-ebml/. But maybe it is easier to meet this problem if you check all the references within the document when there is another version that displays the sections numerically? Or since it looks like there was some automatic fill in of "Section x.x.x.x." when https://datatracker.ietf.org/doc/draft-lhomme-cellar-ebml/ was created should we try to prevent those kind of problems by changes in the original?
I took the second approach in my changes, but I totally understand your point. Should I take every change back that was related to the sections?

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -50,7 +50,7 @@ The VINT\_MARKER serves as a separator between the VINT\_WIDTH and VINT_DATA. Ea
 
 ### VINT_DATA
 
-The VINT\_DATA portion of the Variable Size Integer includes all data that follows (but not including) the VINT\_MARKER until end of the Variable Size Integer whose length is derived from the VINT\_WIDTH. The bits required for the VINT\_WIDTH and the VINT\_MARKER combined use one out of eight bits of the total length of the Variable Size Integer. Thus a Variable Size Integer of 1 octet length supplies 7 bits for VINT\_DATA, a 2 octet length supplies 14 bits for VINT\_DATA, and a 3 octet length supplies 21 bits for VINT\_DATA. If the number of bits required for VINT\_DATA are less than the bit size of VINT\_DATA, then VINT\_DATA may be zero-padded to the left to a size that fits. The VINT\_DATA value MUST be expressed as a big-endian unsigned integer.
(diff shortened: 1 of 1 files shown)


mbunkus:
Thanks for the pull request, but I don't quite agree with all the changes. What's the reason you've removed the articles in front of links to other sections, e.g. saying "See section of expression of range…" instead of "See the section of expression of range…"? It may look OK as an HTML document, but when you create a text document (and the RFCs will be text documents) there won't be hyperlinks, and it will just sound pretty strange.

Additionally I don't agree with another change, but I'll add a line note for that in a moment.

specification.markdown
```diff
@@ -50,7 +50,7 @@ The VINT\_MARKER serves as a separator between the VINT\_WIDTH and VINT_DATA. Ea
...
+The VINT\_DATA portion of the Variable Size Integer includes all data that follows (but not including) the VINT\_MARKER until end of the Variable Size Integer whose length is derived from the VINT\_WIDTH. The bits required for the VINT\_WIDTH and the VINT\_MARKER combined use one out of eight bits of the total length of the Variable Size Integer. Thus a Variable Size Integer of 1 octet length supplies 7 bits for VINT\_DATA, a 2 octet length supplies 14 bits for VINT\_DATA, and a 3 octet length supplies 21 bits for VINT\_DATA. If the number of bits required for VINT\_DATA are less than the bit size of VINT\_DATA, then VINT\_DATA may be zero-padded to the left to a size that fits. The VINT\_DATA value MUST be expressed as a big-endian unsigned integer.
```
mbunkus: I don't really like this change; to me it sounds even more confusing than the original. Please either revert it or think of something more descriptive. Thanks.

marionj1: Again I was just filling in a suggestion by Tim. I will revert it.

robUx4: I agree the `one bit per octet of the total length` (only change I could find) seems better.
Everything else looks fine.

mbunkus:
Thanks :)

dericed:
The "See the Section 1.11.1.4" parts that are seen in the RFC output are a problem with makefile and not with the markdown file itself. I'll file a issue for this.

PR #80, opened 2016-07-30
typos

retokromer:


mbunkus:
Thanks.

specification.markdown
```diff
@@ -10,7 +10,7 @@ EBML uses a simple approach of building Elements upon three pieces of data (tag,
 
 ## Notation and Conventions
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [@!RFC2119](https://tools.ietf.org/html/rfc2119).
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://tools.ietf.org/html/rfc2119).
```
dericed: Just a note that the `@!` is used by [mmark](https://github.com/miekg/mmark/wiki/Syntax#citations) in the process to create the RFC output. We'll somehow need to find a compromise to support the auto-creation of the references section that works for each output.

mbunkus: I wasn't aware of that. Sorry for merging so quickly without waiting for feedback first.

One possibility is to leave those markers in and to include a small script in the build process that removes them (they're very distinct and easy to grab via regexes) when outputting to a format that doesn't use them.

retokromer: Hmm... there SHOULD be a solution ;-) I'll try to find out.

retokromer: It seems that by declaring the `@!` as markdown comment it doesn't work. The best is possibly to revert and keep only the Class D typo/error.

mbunkus:
Reverted.

retokromer:
Thank you! And sorry for the confusion...

PR #83, opened 2016-07-31
[RFC] Explicit the EBML parent of an element

robUx4:
With the level-based definition of an element we rely on the order of which the elements are defined. It also makes it tricky to define elements in a separate specification (think WebM elements, DivX elements, or deprecated elements that won't make it to the RFC).

I think we should set the explicit parent for each element. That means also including the parents of the parent too.
For example ChapterTimeStart: Chapters.EditionEntry.ChapterAtom


specification.markdown
```diff
@@ -223,6 +223,7 @@ Within an EBML Schema the `<element>` uses the following attributes to define an
 | attribute name | required | definition |
 |:---------------|:---------|:-----------|
 | name           | Yes      | The official human-readable name of the EBML Element. The value of the name MUST be in the form of an NCName as defined by the [XML Schema specification](http://www.w3.org/TR/1999/REC-xml-names-19990114/#ns-decl). |
+| parent         | Yes      | The name of the parent EBML Elements. This value MUST be defined with the full hierarchy of EBML Elements separated with a dot. The top EBML Element in the hierarchy being the first in the value. Elements at the top level MUST use the 'none' parent value. Elements that can be found anywhere in the EBML Body or Header MUST use the 'any' parent value. |
```
mbunkus: I like the general idea. I slightly dislike assigning special meaning to ordinary words (`none` and `any`). This prevents any other EBML-based specification from using those words as element names.

What about using symbols? E.g. `parent="-"` for no parent and `parent="*"` for it being allowed in any parent. Alternatively just leave `parent` empty for no parent.

Additionally I'd like to use an established method of specifying a path to an element, e.g. how [XPath's Location Path Expressions](http://www.w3schools.com/xsl/xpath_axes.asp). It's well-known and understood. To use your example from the pull request: `/Chapters/EditionEntry/ChapterAtom`

ablwr: +1 for the XPath syntax both in the location path expression and in the parent assignment definition. Sold on * for any. Not totally sold on - for "none."

robUx4: For the root element (currently known as level 0) it would be logical to have the parent as `/`. An empty value for elements that can be found anywhere (like Void and CRC32) would make sense too.

robUx4:
I updated the patch to use a path and `*` for elements with no specific parent. It's similar to a wildcard for a path.

specification.markdown
```diff
@@ -10,7 +10,7 @@ EBML uses a simple approach of building Elements upon three pieces of data (tag,
 
 ## Notation and Conventions
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [@!RFC2119](https://tools.ietf.org/html/rfc2119).
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://tools.ietf.org/html/rfc2119).
```
mbunkus: Please remove this change; as discussed in #80 this should not have been done in the first place, and I've purposefully reverted it since.

mbunkus:
Great :) I wasn't too happy with `-` for "no parent" either; `/` is much better and in line with XPath.

Other than the invalid removal of `@!` it looks good to me now.

robUx4:
I broadened the interpretation of the `*` so now `Void` is a child f `*` and `CRC32` is a child of `/*`

mbunkus:
:+1:

specification.markdown
```diff
@@ -223,6 +223,7 @@ Within an EBML Schema the `<element>` uses the following attributes to define an
 | attribute name | required | definition |
 |:---------------|:---------|:-----------|
 | name           | Yes      | The official human-readable name of the EBML Element. The value of the name MUST be in the form of an NCName as defined by the [XML Schema specification](http://www.w3.org/TR/1999/REC-xml-names-19990114/#ns-decl). |
+| parent         | Yes      | The path to the parent EBML Element. This path MUST be defined with the full hierarchy of EBML Elements separated with a `/`. The top EBML Element in the path hierarchy being the first in the value. Elements that can be found anywhere inside en EBML Element and its children MUST use the `*` (wildcard) value after the path of the first parent allowed. |
```
dericed: `parent EBML Element` -> `Parent Element`

dericed: Re: 
> This path MUST be defined with the full hierarchy of EBML Elements separated with a `/`.

I'm not sure this is helpful. I don't oppose, but think that Parent Element alone is more flexible and less error-prone.

dericed: Re:
> Elements that can be found anywhere inside en EBML Element and its children MUST use the `*` (wildcard) value after the path of the first parent allowed.

I would prefer to express multiple Parent Elements in this case. Such as, for SimpleTag, `parent="Tag|SimpleTag"`.

Although having multiple values concatenated within an attribute is bad XML form. Perhaps better to use like:

```xml
<element name="SimpleTag" cppname="TagSimple" level="3" id="0x67C8" type="master" minOccurs="1" maxOccurs="unbounded" minver="1" webm="0">
   <parent>Tag</parent>
   <parent>SimpleTag</parent>
   <documentation lang="en">Contains general information about the target.</documentation>
</element>
```

Removed the `recursive` attribute in the example above since the `parent` sub-element implies it.

dericed: The `parent` definition implies that the path is expressed with Element Names but it is not clear enough. For instance from a reading of this I may try to say `0x1946696C/0x6146/0x614E`.

mbunkus: Re:

> I'm not sure this is helpful. I don't oppose, but think that Parent Element alone is more flexible and less error-prone.

I'm not fond of having to specify the full path either, for the same reason. It also annoys the database designer in me as information isn't normalized and minimized.

Re:

> I would prefer to express multiple Parent Elements in this case. Such as, for `SimpleTag`, `parent="Tag|SimpleTag"`.

Me, too. Let's reserve `*` for "can occur anywhere". I have no strong feelings about keeping it an attribute with special syntax or using multiple child nodes for it.

robUx4: OK, I can live with the parent just above. It requires each name throughout the whole DocType(s) to be unique.

The parent="Tag|SimpleTag" is certainly better clearer, should we remove the recursive flag then ? I suppose the Global attribute covers the case of `*` but what parent should these parents have then ? Or do we also remove the `global` attribute and assume no `parent` equals can be found anywhere ?

dericed:
I notice that this PR will conflict with https://github.com/Matroska-Org/ebml-specification/pull/101. I suggest we consider https://github.com/Matroska-Org/ebml-specification/pull/101 first and can then resolve conflicts here after.

robUx4:
I agree. I will update it once it's done.

dericed:
I resolved conflicts so that this is mergable and updated a bit of the vocab in @robUx4's parent attribute definition.

I suppose I still think it would be better to simply have the name of the parent rather than full path to the parent.

dericed:
added a new commit to change from saying the <EBMLSchema>/<element> should be hierarchical to say it should be flat.

specification.markdown
```diff
@@ -217,7 +217,7 @@ An EBML Stream is a file that consists of one or many EBML Documents that are co
...
+Within the `EBML Schema` each `EBML Element` is defined to occur as the `Root Element` or as a `Child Element` to a specified `Parent Element`. For any specified `EBML Element` that is not at level 0, the `Parent Element` refers to the `Master Element` that that `EBML Element` is contained within. For any specified `Master Element` the `Child Element` refers to the `EBML Element` that is immediately contained within that `Master Element`. For any `EBML Element` that is not defined at level 0, the `Parent Element` is identified by the `parent` attribute.
```
retokromer: that that
[The whole sentence sounds strange to me.]

specification.markdown
```diff
@@ -227,7 +227,7 @@ The `EBML Schema` does not itself document the `EBML Header`, but documents all
 
 ### <EBMLSchema> Element
 
-As an XML Document, the `EBML Schema` MUST use `<EBMLSchema>` as the top level element. The `<EBMLSchema>` element MAY contain `<element>` sub-elements.
+As an XML Document, the `EBML Schema` MUST use `<EBMLSchema>` as the top level element. The `<EBMLSchema>` element MSUT contain `<element>` sub-elements.
```
retokromer: MUST

dericed: updated and squashed

specification.markdown
dericed: 'that the' is better?

retokromer: any ... that ... that ... within
In other languages than English I can see simpler formulations.

robUx4: Fixed with 84eb8c47394b694dd4c88ec3c0ea6e7df05da6c6

dericed:
I think this is ready for a new review.

specification.markdown
```diff
@@ -365,29 +371,29 @@ The `type` attribute is OPTIONAL.
...
   maxOccurs="unbounded">
    <documentation lang="en" type="definition">An attached file.
    </documentation>
-   <element name="FileName" level="2" id="0x614E" type="utf-8"
+   <element name="FileName" parent="/Files/File" level="2" id="0x614E" type="utf-8"
    minOccurs="1">
     <documentation lang="en" type="definition">Filename of the attached
     file.</documentation>
    </element>
-   <element name="MimeType" level="2" id="0x464D" type="string"
```
dericed: I prefer to simply say `parent="File"` with the Parent Element Name only and not the full path.

dericed:
I added a commit to drop the `global` attribute since it's now redundant to the `parent` attribute. Is it possible for `parent` to say that `CRC-32` can occur anywhere but not at Level 0?

robUx4:
For `CRC-32` the `/*` syntax means the path has to have `/` and then anything. But maybe that's not good enough. Any element at the root is also `/<something>`.

We could use something like `/<top level element>/*` which would imply it has at least that parent. For now the top level element is single and there's Void that can also be at the top. If we decide one day that EBML can have mutliple top level elements in `EBML Body` that won't work. How about `/<TopLevelElement(s)>/*` ? If the goal is to make the parent easily parseable that's not very good. We may need a specific keyword that says it means every Top Level Element(s) in the Body.

robUx4:
Maybe the simplest parseable format/regular expression would be `/*/*` meaning it has to be at inside another element which can be anything and at any level.

robUx4:
<removed 2 previous comments as I was defining the path to the element, not to the parent>

For `CRC-32` the `/+` would probably be a better parent description. It means it has to be at least an actual element path (with many levels possible). 
To make things more consistent `Void` should probably be defined with a parent of `/*` meaning even `/` qualifies. We should probably specify how these regular expressions should be interpreted. Is there any good standard for that ?

retokromer:
> We should probably specify how these regular expressions should be interpreted.

:+1:

In regex terms, the repetition metacharacter `*` means none to infinite and `+` means 1 to infinite.

robUx4:
Yes but is there a standard for that ? (I haven't looked for it yet)

retokromer:
POSIX is, at least, the _de facto_ standard. I will try to check it, and find out more.

retokromer:
He he he! The `man perlre` on my computer says (in the section Bugs) "This document needs a rewrite that separates the tutorial content from the reference content." I fear we have first to standardise Regex...

dericed:
The xpath spec also defines ?, *, + in a similar way. http://www.w3.org/TR/xpath-31/#id-matching-value

robUx4:
Actually the way the [XPath `SequenceType`](https://www.w3.org/TR/xpath-31/#id-sequencetype-syntax) is described could be modified to fit our needs. Only an `ElementName` would be left open for any name (maybe an ASCII sequence including capital letters and maybe numbers). This is a good starting point.

I also like this section of the [XML specs](https://www.w3.org/TR/REC-xml/#sec-notation) that's even simpler and very readable. It could be expanded a little for our path.

retokromer:
What is the current status of this pull request?

dericed:
I think it needs a review and conflicts resolved

robUx4:
I'm on it

robUx4:
#106 makes more sense

retokromer:
Thank you for clarifying!

PR #102, opened 2016-09-18
docType should not be empty

dericed:


robUx4:
Do we want a max size ?
Maybe someday the doctype may be extended with more subtle uses like extensions and such.

dericed:
I initially had it as `>0`. I could put that back.

dericed:
Ok, back at `>0`.

dericed:
Should this be merged or closed? Can an empty string be `A string that describes and identifies the content of the EBML Body that follows this EBML Header`?

PR #106, opened 2016-09-25
[RFC] replace the level/minOccurs/maxOccurs attributes with a EBNF based path attribute

robUx4:
This deprecates PR #83.

The `EBML Schema` has less attributes but the path needs to be parsed to figure out the hierarchy and occurences. We may need to split the flat/textual element definitions and the `EBML Schema` which can present the data in the way they were before this patch. Each part could be mapped to one section of the EBNF path defined in the textual definition.

I don't know if the XML Schema is harder to parse with the path of with different attributes.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -315,37 +315,47 @@ The `name` attribute is REQUIRED.
 
 #### path
 
-The path defines the allowed storage locations of the `EBML Element` within an `EBML Document`. This path MUST be defined with the full hierarchy of `EBML Elements` separated with a `/`. The top `EBML Element` in the path hierarchy being the first in the value. The syntax of the `path` attribute is defined using this Augmented Backus-Naur Form (ABNF) [@!RFC5234] notation:
+The path defines the allowed storage locations of the `EBML Element` within an `EBML Document`. This path MUST be defined with the full hierarchy of `EBML Elements` separated with a `/`. The top `EBML Element` in the path hierarchy being the first in the value. The syntax of the `path` attribute is defined using this Augmented Backus-Naur Form (ABNF) [@!RFC5234] with the case sensitive update [@!RFC7405] notation:
 
 The `path` attribute is REQUIRED.
 
 ```
-EBMLFullPath          = [EBMLElementOccurence] EBMLReferencePath
-EBMLReferencePath     = EBMLParentPath EBMLElementPath
-EBMLParentPath        = *EBMLPathAtom / AllWildcard
-EBMLElementPath       = EBMLPathAtom / EBMLPathAtomRecursive
-EBMLPathAtom          = PathDelimiter EBMLAtomName
-EBMLPathAtomRecursive = "(1*" EBMLPathAtom ")"
-EBMLAtomName          = 1*EBMLNameChar / AllWildcard
-EBMLNameChar          = [A-Z] / [a-z] / [0-9] / "-" / "."
-AllWildcard           = "any"
-PathDelimiter         = "\"
-MinOccurence          = [0-9]+
-MaxOccurence          = [0-9]+
-EBMLElementOccurence  = [MinOccurence] "*" [MaxOccurence]
+EBMLFullPath            = EBMLElementOccurence "(" EBMLReferencePath ")"
+EBMLReferencePath       = [EBMLParentPath] EBMLElementPath
+EBMLParentPath          = EBMLFixedParent EBMLLastParent
+EBMLFixedParent         = *(EBMLPathAtom)
+EBMLElementPath         = EBMLPathAtom / EBMLPathAtomRecursive
+EBMLPathAtom            = PathDelimiter EBMLAtomName
+EBMLPathAtomRecursive   = "1*(" EBMLPathAtom ")"
+EBMLLastParent          = EBMLPathAtom / EBMLVariableParent
+EBMLVariableParent      = VariableParentOccurence "(\any)"
+EBMLAtomName            = 1*(EBMLNameChar)
+EBMLNameChar            = ALPHA / DIGIT / "-" / "."
+PathDelimiter           = "\"
+AllWildcard             = %s"any"
+EBMLElementOccurence    = [EBMLMinOccurence] "*" [EBMLMaxOccurence]
+EBMLMinOccurence        = 1*DIGIT
+EBMLMaxOccurence        = 1*DIGIT
+VariableParentOccurence = [PathMinOccurence] "*" [PathMaxOccurence]
+PathMinOccurence        = 1*DIGIT
+PathMaxOccurence        = 1*DIGIT
 ```
 
-The `"*"`, `"("` and `")"` symbols MUST be interpreted as they are defined in the ABNF. The `EBMLAtomName` MUST NOT use the reserved word `"any"` corresponding to the `AllWildcard` value.
+The `"*"`, `"("` and `")"` symbols MUST be interpreted as they are defined in the ABNF.
 
 The `EBMLPathAtom` part of the `EBMLElementPath` MUST be equal to the `name` attribute of the `EBML Schema`.
 
-The `MinOccurence` represents the minimum number of occurrences of this `EBML Element` within its `Parent Element`. Each instance of the `Parent Element` MUST contain at least this many instances of this `EBML Element`. If the `EBML Element` has an empty `EBMLParentPath` then `MinOccurence` refers to constaints on the occurrence of the `EBML Element` within the `EBML Document`. If `MinOccurence` is not present then that `EBML Element` is considered to have a `MinOccurence` value of 0. The semantic meaning of `MinOccurence` within an `EBML Schema` is considered analogous to the meaning of `minOccurs` within an `XML Schema`. `EBML Elements` with `MinOccurence` set to "1" that also have a `default` value (see [default](#default)) declared are not REQUIRED to be stored but are REQUIRED to be interpreted, see [Note on the Use of default attributes to define Mandatory EBML Elements](#note-on-the-use-of-default-attributes-to-define-mandatory-ebml-elements). An `EBML Element` defined with a `MinOccurence` value greater than zero is called a `Mandatory EBML Element`.
+The starting `PathDelimiter` of the path corresponds to the root of the `EBML Document`.
 
-The `MaxOccurence` represents the maximum number of occurrences of this `EBML Element` within its `Parent Element`. Each instance of the `Parent Element` MUST contain at most this many instances of this `EBML Element`. If the `EBML Element` has an empty `EBMLParentPath` then `MaxOccurence` refers to constaints on the occurrence of the `EBML Element` within the `EBML Document`. If `MaxOccurence` is not present then that `EBML Element` is considered to have an unbounded `MaxOccurence` value. The semantic meaning of `MaxOccurence` within an `EBML Schema path` is considered analogous to the meaning of `maxOccurs` within an `XML Schema`.
+The `EBMLElementOccurence` part is interpreted as an ABNF Variable Repetition. The repetition amounts correspond to how many times the `EBML Element` can be found in its parent `Parent Element`.
+
+The `EBMLMinOccurence` represents the minimum number of occurrences of this `EBML Element` within its `Parent Element`. Each instance of the `Parent Element` MUST contain at least this many instances of this `EBML Element`. If the `EBML Element` has an empty `EBMLParentPath` then `EBMLMinOccurence` refers to constaints on the occurrence of the `EBML Element` within the `EBML Document`. If `EBMLMinOccurence` is not present then that `EBML Element` is considered to have a `EBMLMinOccurence` value of 0. The semantic meaning of `EBMLMinOccurence` within an `EBML Schema` is considered analogous to the meaning of `minOccurs` within an `XML Schema`. `EBML Elements` with `EBMLMinOccurence` set to "1" that also have a `default` value (see [default](#default)) declared are not REQUIRED to be stored but are REQUIRED to be interpreted, see [Note on the Use of default attributes to define Mandatory EBML Elements](#note-on-the-use-of-default-attributes-to-define-mandatory-ebml-elements). An `EBML Element` defined with a `EBMLMinOccurence` value greater than zero is called a `Mandatory EBML Element`.
+
+The `EBMLMaxOccurence` represents the maximum number of occurrences of this `EBML Element` within its `Parent Element`. Each instance of the `Parent Element` MUST contain at most this many instances of this `EBML Element`. If the `EBML Element` has an empty `EBMLParentPath` then `EBMLMaxOccurence` refers to constaints on the occurrence of the `EBML Element` within the `EBML Document`. If `EBMLMaxOccurence` is not present then that `EBML Element` is considered to have an unbounded `EBMLMaxOccurence` value. The semantic meaning of `EBMLMaxOccurence` within an `EBML Schema path` is considered analogous to the meaning of `maxOccurs` within an `XML Schema`.
+
+The `VariableParentOccurence` part is interpreted as an ABNF Variable Repetition. The repetition amounts correspond to the amount of unspecified `Parent Element` levels there can be between the `EBMLFixedParent` and the actual `EBMLElementPath`.
 
 If the path contains a `EBMLPathAtomRecursive` part, the `EBML Element` can occur within itself recursively (see the [recursive attribute](#recursive)).
-
-If the `EBMLParentPath` is `"*"` then that `EBML Element` can occur anywhere in the `EBML Document` including at the root.
 
 #### id
 
@@ -355,13 +365,13 @@ The `id` attribute is REQUIRED.
 
 #### minOccurs
 
-An integer expressing the minimum number of occurrences of this `EBML Element` within its `Parent Element`. The `minOccurs` value MUST be equal to the `MinOccurence` value of the `path`.
+An integer expressing the minimum number of occurrences of this `EBML Element` within its `Parent Element`. The `minOccurs` value MUST be equal to the `EBMLMinOccurence` value of the `path`.
 
 The `minOccurs` attribute is OPTIONAL. If the `minOccurs` attribute is not present then that `EBML Element` is considered to have a `minOccurs` value of 0.
 
 #### maxOccurs
 
-An integer expressing the maximum number of occurrences of this `EBML Element` within its `Parent Element`. The `maxOccurs` value MUST be equal to the `MaxOccurence` value of the `path`.
+An integer expressing the maximum number of occurrences of this `EBML Element` within its `Parent Element`. The `maxOccurs` value MUST be equal to the `EBMLMaxOccurence` value of the `path`.
 
 The `maxOccurs` attribute is OPTIONAL. If the `maxOccurs` attribute is not present then that `EBML Element` is considered to have a maxOccurs value of 1.
 
@@ -379,7 +389,7 @@ The `size` attribute is OPTIONAL. If the `size` attribute is not present for tha
 
 #### default
 
-If an Element is mandatory (has a `MinOccurence` value greater than zero) but not written within its `Parent Element` or stored as an `Empty Element`, then the `EBML Reader` of the `EBML Document` MUST semantically interpret the `EBML Element` as present with this specified default value for the `EBML Element`. `EBML Elements` that are `Master Elements` MUST NOT declare a `default` value.
+If an Element is mandatory (has a `EBMLMinOccurence` value greater than zero) but not written within its `Parent Element` or stored as an `Empty Element`, then the `EBML Reader` of the `EBML Document` MUST semantically interpret the `EBML Element` as present with this specified default value for the `EBML Element`. `EBML Elements` that are `Master Elements` MUST NOT declare a `default` value.
 
 The `default` attribute is OPTIONAL.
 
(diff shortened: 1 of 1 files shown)


dericed:
"unbounded" means the the element may have unlimited occurrences in its Parent. Though there would be limits imposed by EBMLMaxSizeLength.

retokromer:
Then I would say `{natural_number}` aka infinite.

specification.markdown
```diff
@@ -40,17 +40,17 @@ This document defines specific terms in order to define the format and applicati
...
 
 `Master Element`: The `Master Element` contains zero, one, or many other `EBML Elements`.
 
 `Parent Element`: A relative term to describe the `Master Element` which contains a specified element.
 
-`Root Element`: A mandatory, non-repeating `EBML Element` which occurs within an `EBML Document` at `Level 0` and contains all other `EBML Elements` of the `EBML Document`, excepting the `EBML Header` and optional `Void Elements`.
+`Root Element`: A mandatory, non-repeating `EBML Element` which occurs at the top level of the path hierarchy within an `EBML Body` and contains all other `EBML Elements` of the `EBML Body`, excepting optional `Void Elements`.
 
-`Top-Level Element`: An `EBML Element` defined to only occur at `Level 1` within the `Root Element`.
+`Top-Level Element`: An `EBML Element` defined as occuring within the `Root Element`.
```
dericed: I don't think this is as clear as `Level 1` was. How about `defining as a Child Element of the Root Element`.

specification.markdown
```diff
@@ -231,7 +231,7 @@ The `Date Element` stores an integer in the same format as the `Signed Integer E
...
+The `Master Element` contains zero, one, or many other elements. `EBML Elements` contained within a `Master Element` MUST have their `EBMLParentPath` of their `Element Path` equals to the `EBMLReferencePath` of the `Master Element` `Element Path`. `Element Data` stored within `Master Elements` SHOULD only consist of `EBML Elements` and SHOULD NOT contain any data that is not part of an `EBML Element`. When `EBML` is used in transmission or streaming, data that is not part of an `EBML Element` is permitted to be present within a `Master Element` if `unknownsizeallowed` is enabled within the definition for that `Master Element`. In this case, the `EBML Reader` should skip data until a valid `Element ID` of the same `EBMLParentPath` or the next upper level `Element Path` of the `Master Element` is found. What `Element IDs` are considered valid within a `Master Element` is identified by the `EBML Schema` for that version of the `EBML Document Type`. Any data contained within a `Master Element` that is not part of a `Child Element` MUST be ignored.
```
dericed: `EBMLParentPath` needs a forward reference.

specification.markdown
```diff
@@ -313,35 +313,47 @@ The `name` provides the official human-readable name of the `EBML Element`. The
 
 The `name` attribute is REQUIRED.
 
-#### level
+#### path
 
-The `level` notes at what hierarchical depth or depths the `EBML Element` MUST occur if used within an `EBML Document`. The `Root Element` of an `EBML Document` is at Level 0 and the `Child Elements` that it contains are at Level 1. The `level` MUST be expressed as an integer. `EBML Elements` defined as `global` and `recursive` MAY occur at a level greater than or equal to the defined `level`.
+The path of the `EBML Element`. This path MUST be defined with the full hierarchy of `EBML Elements` separated with a `/`. The top `EBML Element` in the path hierarchy being the first in the value. The syntax of the `path` attribute is defined using this Extended Backus-Naur Form (EBNF) [@!ISO.14977.1996] notation:
```
dericed: Suggested expansion of the first sentence:
> The path is an expression of the allowed storage locations of the `EBML Element` within an `EBML Document`.

dericed: Still wary of mandating `full hierarchy`. Would prefer it to be valid to use either `full hierarchy` or `relative hierarchy` anchoring on a reference to the parent.

dericed: It says `separated with a /` but the examples start with a `/`. The first `/` is not separating but serving as a starting point.

specification.markdown
```diff
@@ -653,16 +626,10 @@ description: The minimum `DocType` version an `EBML Reader` has to support to re
 
 name: CRC-32
 
-level: 1
-
-global: true
+path: `/*/CRC-32(0,1)`
```
dericed: This is different. In the earlier definition we had said `EBML Elements defined as global and recursive MAY occur at a level greater than or equal to the defined level.` so CRC-32 could occur within any Master Element so that CRC-32 had a position of Level 1 or greater.

This `path` implies that CRC-32 can not be at Level 2 but only at Level 1.

Perhaps here we need a `/*//CRC-32(0,1)` where `//` means at any place.

dericed: Ok, ignore the prior CRC-32 comment, I see that `*` implies the any position. I'll move my comments to the section on the meaning of `*`.

specification.markdown
```diff
@@ -415,34 +427,27 @@ The `type` attribute is OPTIONAL.
...
+  <documentation lang="en" type="definition">An attached file.</documentation>
+ </element>
+ <element name="FileName" path="/Files/File/FileName(1,1)" id="0x614E" type="utf-8">
+  <documentation lang="en" type="definition">Filename of the attached file.
+  </documentation>
+ </element>
+ <element name="MimeType" path="/Files/File/MimeType(1,1)" id="0x464D" type="string">
+  <documentation lang="en" type="definition">MIME type of the file.</documentation>
+ </element>
+ <element name="ModificationTimestamp" parent="/Files/File/ModificationTimestamp(1,1)"
```
dericed: parent->path

specification.markdown
```diff
@@ -415,34 +427,27 @@ The `type` attribute is OPTIONAL.
...
+  id="0x4654" type="date">
+  <documentation lang="en" type="definition">Modification timestamp of the file.
+  </documentation>
+ </element>
+ <element name="Data" parent="/Files/File/Data(1,1)" id="0x4664" type="binary">
```
dericed: parent->path

specification.markdown
```diff
@@ -301,7 +301,7 @@ Each `<element>` defines one `EBML Element` through the use of several attribute
 
 The `<element>` nodes contain a description of the meaning and use of the `EBML Element` stored within one or many `<documentation>` sub-elements.
 
-The `<element>` nodes MUST be arranged hierarchically according to the permitted structure of the EBML Document Type. An `<element>` node that defines an `EBML Element` which is a `Child Element` of another `Parent Element` MUST be stored as an immediate sub-element of the `<element>` node that defines the Parent Element. `<element>` nodes that define Level 0 Elements and Global Elements should be sub-elements of `<EBMLSchema>`.
+The `<element>` nodes MUST be arranged hierarchically according to the permitted structure of the EBML Document Type. An `<element>` node that defines an `EBML Element` which is a `Child Element` of another `Parent Element` MUST be stored as an immediate sub-element of the `<element>` node that defines the Parent Element. `<element>` nodes that define Root Level Elements and Global Elements should be sub-elements of `<EBMLSchema>`.
```
dericed: With the intent of the PR, there's no longer a reason to say that `element nodes MUST be arranged hierarchically`. This paragraph can be cut down substantially just to:
> All `<element>` nodes MUST be sub-elements of `<EBMLSchema>`.

Since that is very brief I suggest just adding that sentence onto the prior single-sentence paragraph.

specification.markdown
```diff
@@ -255,21 +255,21 @@ All `EBML Elements` within the `EBML Header` MUST NOT use any `Element ID` with
...
 
-An `EBML Schema` MUST declare exactly one `EBML Element` at Level 0 (referred to as the `Root Element`) that MUST occur exactly once within an `EBML Document`. The `EBML Element` and the `Void Element` MAY also occur at Level 0 but are not considered to be `Root Elements` (see [the definition of the `Void Element`](#void-element)).
+An `EBML Schema` MUST declare exactly one `EBML Element` at Root Level (referred to as the `Root Element`) that MUST occur exactly once within an `EBML ody`. The `Void Element` MAY also occur at Root Level but is not considered to be `Root Elements` (see [the definition of the `Void Element`](#void-element)).
```
dericed: `EBML ody`->`EBML Body`

specification.markdown
```diff
@@ -255,21 +255,21 @@ All `EBML Elements` within the `EBML Header` MUST NOT use any `Element ID` with
...
+All data of an `EBML Document` following the `EBML Header` is the `EBML Body`. The end of the `EBML Body`, as well as the end of the `EBML Document` that contains the `EBML Body`, is considered as whichever comes first: the beginning of a new `EBML Header` at the Root Level or the end of the file. The `EBML Body` MUST consist only of `EBML Elements` and MUST NOT contain any data that is not part of an `EBML Element`. This document defines precisely what `EBML Elements` are to be used within the `EBML Header`, but does not name or define what `EBML Elements` are to be used within the `EBML Body`. The definition of what `EBML Elements` are to be used within the `EBML Body` is defined by an `EBML Schema`.
```
dericed: Suggesting to add `Root Level` to the terminology section as now the term appears in a few places.

specification.markdown
```diff
@@ -313,35 +313,47 @@ The `name` provides the official human-readable name of the `EBML Element`. The
...
+EBMLElementPath       ::= EBMLPathAtom | EBMLPathAtomRecursive
+EBMLPathAtom          ::= "/", EBMLAtomName
+EBMLPathAtomRecursive ::= "(", EBMLPathAtom, ")", "+"
+EBMLAtomName          ::= (EBMLNameChar)+ | AllWildcard | AtLeastOnce
+EBMLNameChar          ::= [A-Z] | [a-z] | [0-9] | "-" | "."
+AllWildcard           ::= "*"
+AtLeastOnce           ::= "+"
+EBMLElementOccurence  ::= "(", [MinOccurence], ",", [MaxOccurence] ")" 
+MinOccurence          ::= [0-9]+
+MaxOccurence          ::= [0-9]+ | "unbounded"
```
dericed: what about maxOccurs=`identical` to permit multiple copies for resiliency that must all be equal?

specification.markdown
dericed: Suggest doing so by adding
> (see [the section on the `EBML Path`](#path))

to the end of the first sentence

specification.markdown
```diff
@@ -255,21 +255,21 @@ All `EBML Elements` within the `EBML Header` MUST NOT use any `Element ID` with
...
+For any specified `EBML Element` that is not at Root Level, the `Parent Element` refers to the `Master Element` in which that `EBML Element` is contained. For any specified `Master Element` the `Child Element` refers to the `EBML Element` that is immediately contained within that `Master Element`. For any `EBML Element` that is not defined at Root Level, the `Parent Element` is identified by the preceding `<element>` node which has the `EBMLReferencePath` equals to the `EBMLParentPath` of that `EBML Element` `Element Path`. The only exception for this rule are `Global Elements` which MAY occur within any `Parent Element` within the restriction of the `level` declaration of `Global Element`.
```
dericed: > the `Parent Element` is identified by the preceding `<element>` node

is from back when that's how specdata.xml was interpreted (counted back elements to a lower level to find the parent).

I suggest dropping the word `preceding`.

dericed: Also consider deleting the last sentence.

specification.markdown
dericed: Why change `EBML Document` to `EBML Body`? Can an `EBML Document` have more than one `EBML Body`?

specification.markdown
dericed: The `name` attribute is defined as a valid `NCName` but here it implies that EBMLNameChar is `[A-Z] | [a-z] | [0-9] | "-" | "."`. These should be synced, so that `name` is also constrained to `[A-Z] | [a-z] | [0-9] | "-" | "."`. Also it should assert somewhere that `name` and `EBMLAtomName` MUST be the same.

specification.markdown
```diff
@@ -313,35 +313,47 @@ The `name` provides the official human-readable name of the `EBML Element`. The
...
+The `"*"`, `"+"`, `"("`, `")"`, `"["` and `"]"` symbols MUST be interpreted as they are defined in the EBNF.
 
-A boolean to express if an `EBML Element` MUST occur at its defined level or may occur within any `Parent Element`.
+The `MinOccurence` represent the minimum number of occurrences of this `EBML Element` within its `Parent Element`. Each instance of the `Parent Element` MUST contain at least this many instances of this `EBML Element`. If the `EBML Element` has an empty `EBMLParentPath` then `MinOccurence` refers to constaints on the occurrence of the `EBML Element` within the `EBML Document`. The semantic meaning of `MinOccurence` within an `EBML Schema` is considered analogous to the meaning of `minOccurs` within an `XML Schema`. `EBML Elements` with `MinOccurence` set to "1" that also have a `default` value (see [default](#default)) declared are not REQUIRED to be stored but are REQUIRED to be interpreted, see [Note on the Use of default attributes to define Mandatory EBML Elements](#note-on-the-use-of-default-attributes-to-define-mandatory-ebml-elements). An `EBML Element` defined with a `MinOccurence` value greater than zero is called a `Mandatory EBML Element`.
```
dericed: `The \`MinOccurence\` represents`

specification.markdown
```diff
@@ -313,35 +313,47 @@ The `name` provides the official human-readable name of the `EBML Element`. The
...
 
-The `global` attribute is OPTIONAL. If the `global` attribute is not present then that `EBML Element` is to be considered not global.
+If `MinOccurence` is not present then that `EBML Element` is considered to have a `MinOccurence` value of 0.
```
dericed: I'm not sure if I understand EBNF enough to understand, but does the definition above allow `MinOccurence` to not be present?

specification.markdown
```diff
@@ -313,35 +313,47 @@ The `name` provides the official human-readable name of the `EBML Element`. The
...
 
-The `Element ID` encoded as a `Variable Size Integer` expressed in hexadecimal notation prefixed by a `0x` that is read and stored in big-endian order. To reduce the risk of false positives while parsing `EBML Streams`, the `Element IDs` of the `Root Element` and `Top-Level Elements` SHOULD be at least 4 octets in length. `Element IDs` defined for use at Level 0 or Level 1 MAY use shorter octet lengths to facilitate padding and optimize edits to `EBML Documents`; for instance, the `Void Element` uses an `Element ID` with a one octet length to allow its usage in more writing and editing scenarios.
+If `MaxOccurence` is not present then that `EBML Element` is considered to have a `MaxOccurence` value of 1.
 
-The `id` attribute is REQUIRED.
-
-#### minOccurs
+The `"()+"` group of the `EBMLElementPath` specify that the `EBML Element` can occur within itself recursively (see [Identically Recurring Element](#identically-recurring-elements)).
```
dericed: This seems like a different meaning of `Identically Recurring Element`. I thought an `Identically Recurring Element` would be like copies of the `Tracks` element occurring every once in a while within a Matroska stream, without recursive placement.

specification.markdown
retokromer: How do you define `identical` in EBNF? Feasible, but should be added.

robUx4: `::=` vs `=` you're right @retokromer

robUx4: I do not understand what `identical` would mean as a maximum occurrence. Identical to what ? Or you mean repetition of the same element ? IMO that would be a different field than minimum and maximum occurrence.

specification.markdown
robUx4: I don't want the relative hierarchy because it makes parsing harder and impose a non flat hierarchy in the XML version which contradicts the (possible) possibility of allowing multiple pathes for the same element.

robUx4: I leave the separator as-is for now as we're going to define an actual root before that. Either `body`, `header` or `(body | header)` for `CRC-32` and `Void`

specification.markdown
robUx4: No my initial thought was that the Schema was describing the `EBML Body` only but now I realize that the `EBML Header` elements are also defined the same way.

specification.markdown
robUx4: Yes, that's the [] around it which means optional.

https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_Form

robUx4:
I was looking at the french EBNF page on Wikipedia. On the English one there's more info https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_Form

Especially the repitition is actually normalized (YAY!).

specification.markdown
robUx4: Indeed, the notation doesn't say anything about identical elements.

robUx4:
Reading more about the repetition, that's not the kind we want. It doesn't allow minimum and maximum.

[ABNF](https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_Form) on the other hand does. It's also known as [RFC 5234](http://www.ietf.org/rfc/rfc5234.txt) so probably a better fit for us. I will investigate.

specification.markdown
```diff
@@ -255,21 +257,21 @@ All `EBML Elements` within the `EBML Header` MUST NOT use any `Element ID` with
...
+For any specified `EBML Element` that is not at `Root Level`, the `Parent Element` refers to the `Master Element` in which that `EBML Element` is contained. For any specified `Master Element` the `Child Element` refers to the `EBML Element` that is immediately contained within that `Master Element`. For any `EBML Element` that is not defined at `Root Level`, the `Parent Element` is identified by the `<element>` node which has the `EBMLReferencePath` equals to the `EBMLParentPath` of that `EBML Element` `Element Path`. The only exception for this rule are `Global Elements` which MAY occur within any `Parent Element` within the restriction of the `level` declaration of `Global Element`.
```
dericed: > node which has the `EBMLReferencePath` equals
should be

> node which has the `EBMLReferencePath` equal

dericed: > The only exception for this rule are `Global Elements` which MAY occur within any `Parent Element` within the restriction of the `level` declaration of `Global Element`.

This includes a reference to `level`. Also we have a `Global Element which MAY occur without a `Parent Element`, such as a `VOID` in between the `EBML Header` and `EBML Body`.

specification.markdown
```diff
@@ -673,9 +692,7 @@ description: The `CRC-32 Element` contains a 32-bit Cyclic Redundancy Check valu
 
 name: Void
 
-level: 0
-
-global: true
+path: `*(*any\)Void`
```
dericed: Should this be `*(*any)\Void` ?

specification.markdown
dericed: `identical` is when the element MAY repeat within the Parent but that each repetition MUST be identical (like sending the Tracks element multiple times within a live Matroska stream).

specification.markdown
```diff
@@ -315,37 +315,47 @@ The `name` attribute is REQUIRED.
...
+EBMLElementOccurence    = [EBMLMinOccurence] "*" [EBMLMaxOccurence]
+EBMLMinOccurence        = 1*DIGIT
+EBMLMaxOccurence        = 1*DIGIT
+VariableParentOccurence = [PathMinOccurence] "*" [PathMaxOccurence]
+PathMinOccurence        = 1*DIGIT
+PathMaxOccurence        = 1*DIGIT
 ```
 
-The `"*"`, `"("` and `")"` symbols MUST be interpreted as they are defined in the ABNF. The `EBMLAtomName` MUST NOT use the reserved word `"any"` corresponding to the `AllWildcard` value.
+The `"*"`, `"("` and `")"` symbols MUST be interpreted as they are defined in the ABNF.
```
dericed: Why is the warning about the reserved word `any` removed. If an `EBML DocType` defined `any` as an `Element Name` would there be a confusion in interpretting the path.

specification.markdown
```diff
@@ -355,13 +365,13 @@ The `id` attribute is REQUIRED.
...
 
-An integer expressing the minimum number of occurrences of this `EBML Element` within its `Parent Element`. The `minOccurs` value MUST be equal to the `MinOccurence` value of the `path`.
+An integer expressing the minimum number of occurrences of this `EBML Element` within its `Parent Element`. The `minOccurs` value MUST be equal to the `EBMLMinOccurence` value of the `path`.
 
 The `minOccurs` attribute is OPTIONAL. If the `minOccurs` attribute is not present then that `EBML Element` is considered to have a `minOccurs` value of 0.
 
 #### maxOccurs
 
-An integer expressing the maximum number of occurrences of this `EBML Element` within its `Parent Element`. The `maxOccurs` value MUST be equal to the `MaxOccurence` value of the `path`.
+An integer expressing the maximum number of occurrences of this `EBML Element` within its `Parent Element`. The `maxOccurs` value MUST be equal to the `EBMLMaxOccurence` value of the `path`.
```
dericed: If `maxOccurs` MUST equal `EBMLMaxOccurence` does that infer that `maxOccurs` is mandatory?

specification.markdown
dericed: The Header elements are defined the same way, but should they be part of an EBML Schema? In XML the XML Declaration is not part of the XML Schema.

I think it's fine if the EBML Header is included in the EBML Schema but that means we'd have to mandate the EBML Schemas include an exact definition of the EBML Header elements. Opinion?

specification.markdown
```diff
@@ -614,7 +644,7 @@ description: A string that describes and identifies the content of the `EBML Bod
 
 name: `DocTypeVersion`
 
-level: 1
+path: `1*1(\EBML\DocTypeVersion)`
```
retokromer: Is the 'range: -' on line 655 necessary?

dericed:
There's a few comments above for consideration [remaining mentions of level](https://github.com/Matroska-Org/ebml-specification/pull/106#discussion_r81473273) and handling of [Identically Repeating Elements](https://github.com/Matroska-Org/ebml-specification/pull/106#discussion_r81473273). Should we move these into issues and accept the PR? I prefer not to have a PR open too long.

retokromer:
+1 for moving on and closing this one.

PR #109, opened 2016-10-10
crc recommendation only makes sense with master elements

dericed:
also removes any confusion when the definition of `Top-Level Element`
includes non `Master Elements` such as `Void`


mbunkus:
Yep, the change makes sense. Thanks.

PR #111, opened 2016-10-16
clarify the EBML version range and current value

robUx4:
IMO restricting the value to only 1 hurts forward compatibility of parsers.
Also we MUST say here that the current version is 1.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -556,13 +556,13 @@ minOccurs: 1
 
 maxOccurs: 1
 
-range:  1
+range:  not 0
 
 default: 1
 
 type: Unsigned Integer
 
-description: The version of `EBML Writer` used to create the `EBML Document`.
+description: The version of `EBML` specifications used to create the `EBML Document`. The version of EBML defined in this document is 1, so `EBMLVersion` SHOULD be 1.
 
 ### EBMLReadVersion Element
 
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -556,13 +556,13 @@ minOccurs: 1
 
 maxOccurs: 1
 
-range:  1
+range:  not 0
```
dericed: As noted on Cellar, only version 1 is defined, so any other value would be an undefined implementation. I suggest to leave as `range: 1` and when/if an EBML specification of version 2 is made then this is changed to `range: 1-2`.

specification.markdown
```diff
@@ -556,13 +556,13 @@ minOccurs: 1
...
 
 default: 1
 
 type: Unsigned Integer
 
-description: The version of `EBML Writer` used to create the `EBML Document`.
+description: The version of `EBML` specifications used to create the `EBML Document`. The version of EBML defined in this document is 1, so `EBMLVersion` SHOULD be 1.
```
dericed: LGTM

specification.markdown
robUx4: The text already says 

> `EBMLVersion` SHOULD be 1.

It's a soft requirement because one day value 2 may be valid.

dericed:
I'm convinced by Steve's comments in https://mailarchive.ietf.org/arch/msg/cellar/XNkM0sR6WwRO2oBdh6U9d-vYuGI. +1

retokromer:
+1

PR #115, opened 2016-10-18
XML Schema for EBML Schema

dericed:
allows validation of EBML Schemas

### DIFF
--- a/EBMLSchema.xsd
+++ b/EBMLSchema.xsd
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xsd:schema xmlns="https://ietf.org/cellar/ebml" targetNamespace="https://ietf.org/cellar/ebml" xmlns:xsd="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified" version="01">
+  <xsd:element name="EBMLSchema" type="EBMLSchemaType"/>
+  <xsd:complexType name="EBMLSchemaType">
+    <xsd:sequence>
+      <xsd:element name="element" type="elementType" minOccurs="0" maxOccurs="unbounded"/>
+    </xsd:sequence>
+    <xsd:attribute name="docType" use="required"/>
+    <xsd:attribute name="version" use="required"/>
+  </xsd:complexType>
+  <xsd:complexType name="elementType">
+    <xsd:sequence>
+      <xsd:element name="documentation" type="documentationType" minOccurs="0" maxOccurs="unbounded"/>
+    </xsd:sequence>
+    <xsd:attribute name="name" use="required"/>
+    <xsd:attribute name="path" use="required"/>
+    <xsd:attribute name="id" use="required"/>
+    <xsd:attribute name="minOccurs" default="0"/>
+    <xsd:attribute name="maxOccurs" default="1"/>
+    <xsd:attribute name="range"/>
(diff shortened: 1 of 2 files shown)


EBMLSchema.xsd
```diff
@@ -0,0 +1,45 @@
...
+<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified" version="01">
+  <xsd:element name="EBMLSchema" type="EBMLSchemaType"/>
+  <xsd:complexType name="EBMLSchemaType">
+    <xsd:annotation>
+      <xsd:documentation/>
+    </xsd:annotation>
+    <xsd:sequence>
+      <xsd:element name="element" type="elementType" minOccurs="0" maxOccurs="unbounded"/>
+    </xsd:sequence>
+    <xsd:attribute name="docType" use="required" />
```
retokromer: Cosmetic: <xsd:attribute name="docType" use="required"/>
(no space before closing empty tag)

EBMLSchema.xsd
```diff
@@ -0,0 +1,45 @@
...
+    <xsd:attribute name="version" use="required" />
```
retokromer: ditto

dericed:
I forced an update to respond to @retokromer's comments. I also removed the <xsd:documentation> elements for now. These would be to offer definitions of the elements of the EBML Schema, but I'd rather wait or just reference the EBML specification to learn about what the elements means.

Other issues with this:
- Matroska's ebml_schema.xml uses local attributes (`cppname` and `webm`) we need to allow them or remove them.
- Should EBMLSchema.xsd have a namespace? If so, what?

EBMLSchema.xsd
```diff
@@ -0,0 +1,36 @@
...
+  </xsd:complexType>
+  <xsd:complexType name="elementType">
+    <xsd:sequence>
+      <xsd:element name="documentation" type="documentationType" minOccurs="0" maxOccurs="unbounded"/>
+    </xsd:sequence>
+    <xsd:attribute name="name" use="required"/>
+    <xsd:attribute name="path" use="required"/>
+    <xsd:attribute name="id" use="required"/>
+    <xsd:attribute name="minOccurs" default="0"/>
+    <xsd:attribute name="maxOccurs" default="1"/>
```
robUx4: With this [PR](https://github.com/Matroska-Org/ebml-specification/pull/124) the default for `maxOccurs` is like `EBMLMaxOccurrence` it's infinite.

retokromer:
@dericed In the old days to have a namespace was considered good practice. I didn't follow any discussion about pros and cons of this. Intuitively I would answer yes, and I would chose «something» including ietf and cellar, like `http://ietf.org/cellar/ebml`.

dericed:
Added a namespace based on @retokromer's suggestion, but with an `https` version.

dericed:
If this PR is accepted, I'll update the ebml_matroska.xml file in the matroska spec repo.

retokromer:
LGTM

mbunkus:
LGTM, too. Maybe you could squash both commits and remove the `[WIP]` from the commit message; it might be confusing to let it in.

dericed:
Updated to @mbunkus's recommendations. Please re-review.

retokromer:
LGTM

PR #121, opened 2016-10-23
Spelling and grammar fixes, wording change for spec.md

bastik-tor:
A bit more detailed (mostly individual commits) in case some changes are not supposed to be made.

The last one turns "constaints" into "constraints", which is not obvious the way the diff is presented by Github (at least for me).


dericed:
The commits messages could be more descriptively named but the content all looks good to me. Feasibly these commits could all be squashed to a single commit named 'typos'.

retokromer:
LGTM

bastik-tor:
> The commits messages could be more descriptively named

Yes, I think I should have named them each and not let the default stand as Github put it.

> commits could all be squashed to a single commit named 'typos'.

I'm ignorant on how that is handled. (e.g. cherry-picking just a few and apply other changes that make mine irrelevant or whatever.) (Unless of course there would be something for me to change to provide a better work-flow.)

dericed:
For squashing commits, I'll usually use `git rebase -i HEAD~6` (picking 6 since it would then offer to rebase the past 6 commits, size of your PR). You could then change `pick` to `squash` on commits that you want to squash (in this case, leave first as `pick` and subsequent commits as `squash`). After that you'll have an opportunity to revise the new consolidated commit message and then would need to `git push --force`.

robUx4:
I cherry picked the changes from your branch and fixed the issues. It's now in master with your name on it.
Thanks !

PR #123, opened 2016-10-23
partial revert of

dericed:
https://github.com/Matroska-Org/ebml-specification/commit/7b966f6efc3160
5a7aba61089c98647d085fdea0#diff-01f3b0f2fbf3e44327e6072f64ddc8a0R55 as
mentioned
https://github.com/Matroska-Org/matroska-specification/pull/62#issuecomm
ent-255610832


robUx4:
Makes sense

PR #124, opened 2016-10-30
get rid of unbounded maxOccurs, just use an integer

robUx4:
As discussed on CELLAR unbounded is no longer a valid value. It should become the norm if the value is missing.


dericed:
LGTM

PR #125, opened 2016-11-01
All parents of unknownsizelement MUST also have unknownsizelement

robUx4:


retokromer:
+1

dericed:
suggested re-wording (since we discuss the definition of the EBML Element and not the EBML Element itself I prefer to use the word `define` rather than `has`:
```
An `EBML Element` that is defined with an `unknownsizeallowed` attribute set to 1 MUST also have the `unknownsizeallowed` attribute of its `Parent Element`  set to 1.
```

robUx4:
I updated the PR with @dericed rewording

dericed:
Minor new rewording:
> An `EBML Element` that is defined with an `unknownsizeallowed` attribute set to `1` MUST also have the `unknownsizeallowed` attribute of its `Parent Element`  (if any) set to `1`.

retokromer:
LGTM

mbunkus:
LGTM, too.

robUx4:
I didn't like the `(if any)`. It MUST be set and that's not the default so it must be set anyway.

robUx4:
The backtick of the `1` could be another patch that may be done for other ones as well.

Issue #95, opened 2016-09-12
clarify meaning of default if minOccurs>1


dericed:
Suppose an Element has minOccurs=2 and a default value. Should this be allowed?
Either:
1. the EBML Reader should insert enough copies of the Element to satisfy the minOccurs with default values as needed. Or.
2. We could say that 'default' values are only valid when minOccurs equals 1.

I prefer 2.

retokromer:
I also prefer 2.

mbunkus:
Currently we don't have a case of `minOccurs > 1`, and it hadn't been our intention to allow that in the first place. Remember, `minOccurs=1` was originally the `mandatory` flag.

I don't see a case where `minOccurs > 1` makes sense, but then again, we allowed it…

Oh well. I'm either for disallowing `minOccurs > 1` or for your option 2.

robUx4:
`minOccurs` comes from the analogy with XML. But XML has no default value like we do. Maybe we should bring back the `mandatory` flag and explain it's similar to `minOccurs = 0|1` in XML ?

dericed:
XML does have default values like EBML, but they are only for attributes which may only occur once within an element. It seems a bit controversial since the XML Document is not semantically understood alone but needs the XML Schema to be interpreted.

Although we have no `minOccurs>1` use case, I don't suppose we should force other EBML adopters to have this limit. There is some writing about the zero-one-infinity rule that explains design exceptions. If someone makes an EBML Document Type to describe triangles a `minOccurs=3` may make sense.

I think saying 'default' values are only valid when `minOccurs=1` is the solution most analogous to XML since XML only uses defaults for attributes which infer that more than one occurrence is not possible.

robUx4:
I don't really see a good case for minOccurs=2 or more. Is it ever used in
XML? Maybe if we could think of some cases (not Matroska related) we would
know if a default value makes sense or not. I don't think forcing a
restriction just because we can is a good reason to add this limitation.

Maybe saving a matrix in a linear way could be a good example of minOccurs
= maxOccurs = 9. A default value of 0 would make sense too. The problem is
when you don't write an element because it's default, is it at the
beginning or the end?
Another ensemble of X elements not ordered could be stored linearly too. If
you write only Y elements that means the other X-Y are the default values.
That could save a lot of space in scientific samples.

retokromer:
Yes, you may have `minOccurs=2` in a valid XML.
[...]

PR #129, opened 2017-01-08
defaut is not valid if minOccurs>1

dericed:
intended to resolve https://github.com/Matroska-Org/ebml-specification/issues/95


mbunkus:
I'm fine with the change.

PR #131, opened 2017-02-19
Clean Makefile

robUx4:
A little bonus, you can run make in parallel (make -j2) to have both outputs faster.
The dependencies are handled better.


retokromer:
LGTM

mbunkus:
LGTM

Issue #94, opened 2016-09-11
review method to determine end of unknown-size element


dericed:
Tim asked in cellar review:

> If maxOccurs is a positive integer, and maxOccurs instances of the element have already been encountered within a Master Element with an unknown size, is the another instance of this element considered "...the beginning of the next element that is not a valid sub-element of that Master-element”?

For instance if Segment is unknown sized and knowing the Attachments can only occur once within the Segment, then with this structure:

``` xml
<Segment>
<Attachments/>
<Attachments/>
```

Is the Segment expected to end at when the second occurrence of <Attachments> is read, since a 2nd expression of Attachments "is not a valid sub-element of that Master-element"?

mbunkus:
Ugh, that's ugly. At the moment I'm leaning slightly in the direction of saying that no, this doesn't signal the end of the unknown-sized parent. Solely so that consumers aren't required to track the times a child element has already occurred within its parent.

I'd rather read "…a valid sub-element…" solely from the perspective of whether an element is generally allowed as a child of that parent element. Though I admit this is a bit inconsistent.

robUx4:
No it doesn't mark the end of the unknown sized parent. The semantic level has priority over the number of elements that are valid. Although it's not written that way, the only ways to end an unknown size element is the end of the file, or an element of the same level or upper level.

dericed:
i rewrote this to:

> The end of an `Unknown-Sized Element` is determined by whichever comes first: the end of the file or the beginning of the next `EBML Element`, defined by this document or the corresponding `EBML Schema`, that is not independently valid as Descendent Element of the `Unknown-Sized Element`.

as I think that clarifies the Segment/Attachment/Attachment example above.

dericed:
closing per https://github.com/Matroska-Org/ebml-specification/issues/133

PR #133, opened 2017-02-24
update language of end of unknown element

dericed:
+ define descendant elements


dericed:
intended to fix https://github.com/Matroska-Org/ebml-specification/issues/94

specification.markdown
```diff
@@ -165,7 +167,7 @@ The `Element Data Size` expresses the length in octets of `Element Data`. The `E
...
...o one is reserved as an indicator that the size of the `EBML Element` is unknown. The only reserved value for the `VINT_DATA` of `Element Data Size` is all bits set to one. An `EBML Element` with an unknown `Element Data Size` is referred to as an `Unknown-Sized Element`. Only `Master Elements` SHALL be `Unknown-Sized Elements`. `Master Elements` MUST NOT use an unknown size unless the `unknownsizeallowed` attribute of their `EBML Schema` is set to true (see [the section on the `unknownsizeallowed` attribute](#unknownsizeallowed)). The use of `Unknown-Sized Elements` allows for an `EBML Element` to be written and read before the size of the `EBML Element` is known. `Unknown-Sized Element` MUST NOT be used or defined unnecessarily; however if the `Element Data Size` is not known before the `Element Data` is written, such as in some cases of data streaming, then `Unknown-Sized Elements` MAY be used. The end of an `Unknown-Sized Element` is determined by whichever comes first: the end of the file or the beginning of the next `EBML Element`, defined by this document or the corresponding `EBML Schema`, that is not independently valid as Descendent Element of the `Unknown-Sized Element`.
```
robUx4: Should be back ticked here.

robUx4:
Any reason you picked `Descendent` instead of `Child` ? As elements within a `Master` they might be called `Slave` but I think the master/slave metaphor is not considered insensitive.

Also my spell checker says it should be `Descendant`.

dericed:
Descendant and Child are common in XML to reference relative element relationships.

For instance in:
```xml
<a>
  <b>
    <c/>
  </b>
</a>
```

<b> is a Child Element of <a>, but <c> is not. <b> and <c> are both Descendant Elements of <a>.

Since we had the phrase:
> The end of an `Unknown-Sized Element` is determined by the beginning of the next element, defined by this document or the corresponding `EBML Schema`, that is not a valid sub-element of the `Unknown-Sized Element`.

I wanted to make it clear that if <a> is unknown sized, that the start of <c> does not imply that it is the end of <a>, even though <c> may not be a valid Child of <a>.

dericed:
changed `Descendent` to `Descendant`, and backticked `Descendant Element`

Issue #90, opened 2016-09-05
why are Element IDs with all 0 or all 1 in VINT_DATA reserved?


dericed:
On the [CELLAR list](https://mailarchive.ietf.org/arch/msg/cellar/Mfj44tK1gyfqjU3Uu1tI5fMovtk) Tim asked:

> What is an implementation supposed to do if it encounters a file that violates this restriction? What is the purpose of reserving the various all-ones Element IDs? Or the zero Element ID, for that matter?

which regards the current line in the spec:

> "The VINT_DATA component of the Element ID MUST NOT be set to either all zero values or all one values."

I don't know the answer to this. @robUx4, @mbunkus, do you know the answer to the questions above?

mbunkus:
The easier case is all bits set to 0. This means that the first byte of a potential ID is 0x00, which in turn means that there's no `VINT_MARKER` present. It violates that `MUST` in the definition of `VINT_MARKER`.

Now we're back in "junk" territory. I'm very reluctant to allow consumers to reject files with such invalid data, just as I've said in #92. How do other file formats handle this?

About all bits set to 1: I don't remember why we've disallowed this. I'd say to treat it the same as 0x00, whatever we decide to do with that.

dericed:
> The easier case is all bits set to 0. This means that the first byte of a potential ID is 0x00, which in turn means that there's no `VINT_MARKER` present. It violates that `MUST ` in the definition of `VINT_MARKER`.

No the issue is all 0 as VINT_DATA, such as an Element ID of 0x80, which is currently invalid.

dericed:
Also the first byte of a potential ID is allowed to be 0x00 if the EBMLMaxIDLength is greater than 8.

robUx4:
I'm not sure EBMLMaxIDLength above 8 bytes should be allowed. That makes parsing EBML data a lot more hazardous. 8 bytes ought to be enough for anybody ;)

robUx4:
Historically the boundary values have been reserved in case we need to extend things in an unexpected way. Luckily we never had to use any of these boundary values. The all zero value could have certainly be used for Void.

I really don't know if it's better to mark them as reserved (valid but no meaning yet) or invalid (data are assumed corrupted there).

mbunkus:
I'd prefer reserved.

PR #134, opened 2017-02-24
Say that all-0 or all-1 vint-data must be ignored on read

dericed:
intended to fix
https://github.com/Matroska-Org/ebml-specification/issues/90


robUx4:
I suppose it makes it forward compatible. In any case it's needed in the future, these values won't be ignored and make sense for such readers.

It doesn't say if it's valid or not though. Depending on the answer program that try to detect and fix errors in an EBML Document will have to behave differently. If the goal is that it may be used in the future (and ignored now) that means they are valid values. That would be in the same spirit of the original "reserve" state.

Should we clarify this ?

dericed:
I think `MUST NOT be ... written` is similar to saying invalid, though we could clarify that.

mbunkus:
I'd prefer to say that such elements are reserved in case we do want to make use of them at a later date. Requiring readers to ignore them would be enough for me.

robUx4:
How about:
```Any `Element ID` with the `VINT_DATA` component set as all zero values or all one values MUST be ignored and MUST NOT be considered an error in the `EBML Document`.```

dericed:
updated according to @robUx4's recommendation. please review.

Issue #88, opened 2016-09-05
define EBML Reader & Writer and define its expected behavior


dericed:
Based on [Tim's comment](https://mailarchive.ietf.org/arch/msg/cellar/Mfj44tK1gyfqjU3Uu1tI5fMovtk):

> A number of the things listed as "attacks" are simply things you said someone writing an EBML Document MUST NOT do (or not doing things you said they MUST do). But in none of those cases do you say what is expected of someone reading a document containing these attacks.

@mbunkus, @robUx4, ideas for such a section? There is a bit of similar language at https://github.com/Matroska-Org/matroska-test-files/blob/master/readme.md but I don't think we have such language for EBML Readers yet.

dericed:
This section should note what an EBML Reader should do when encountering an invalid VINT_WIDTH. For instance if a Master Element contains 0b0000000000000000.

dericed:
also from Tim's review: 
> What are you supposed to do if attempting to read a length (e.g., of a String, which has no defined maximum length) and the entire rest of the file consists of zero bytes?

dericed:
Should also define what to do with invalid Element IDs. For instance if the Element Data of a Master Element starts with 0x11111111, this is an invalid Element ID, what should the reader do? The next octet(s) could depict an Element Data Size so that it could be known how much to skip but with an invalid Element ID, I wouldn't trust the next data to be a VINT of Element Data Size.

dericed:
Also what should an EBML Reader do if encountering an Element with unknown size when it is not permitted, such as with a string or integer?

dericed:
Should define behavior for EBML Reader with excessive lengths of strings. For instance a string element in Matroska may have a length of ~72 Petabytes. Perhaps it is unreasonable to expect that the EBML Reader handle this appropriately.

dericed:
As Tim wrote regarding "A String Element may declare any length from zero to 'VINTMAX'."
> MAY? VINTMAX is not a single value. Its value depends on the octet width (length?) of the Element Data Size. Is there any restriction on this octet width? If not, I would just phrase this as, "A String Element can have any length, including zero." However, it is obvious that no implementation will be able to support strings of completely arbitrarily length. Is there some required minimum that decoders are required to handle? When is a 32-bit system with no more than 4 GB of RAM justified in rejecting a string as too large? How about an embedded media player in someone's television? How would someone writing an EBML document know when writing a string if all decoders will be able to handle it?

[...]

PR #135, opened 2017-02-24
start defining ebml reader

dericed:
partially resolves https://github.com/Matroska-Org/ebml-specification/issues/88


robUx4:
LGTM.
Suprising it was backticked but not defined.

PR #136, opened 2017-02-24
backtick all locally defined terms

dericed:
excepting in table headers and section headers

fixes https://github.com/Matroska-Org/ebml-specification/issues/100


robUx4:
LGTM

Issue #44, opened 2016-01-03
clarify the role of null bytes in string/utf elements


dericed:
Early discussion on this here: http://lists.matroska.org/pipermail/matroska-devel/2015-July/004730.html

dericed:
resolved by https://github.com/Matroska-Org/ebml-specification/pull/141

PR #141, opened 2017-04-29
section on terminating elements

dericed:
intended to resolve
https://github.com/Matroska-Org/ebml-specification/issues/44

This moves to info on terminating elements to a new section (to reduce redundancy between the string and utf-8 element descriptions).

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -223,13 +223,13 @@ A `Float Element` stores a floating-point number as defined in [@!IEEE.754.1985]
 
 A `String Element` MUST declare a length in octets from zero to `VINTMAX`. If the `EBML Element` is not defined to have a `default` value, then a `String Element` with a zero-octet length represents an empty string.
 
-A `String Element` MUST either be empty (zero-length) or contain printable ASCII characters [@!RFC0020] in the range of `0x20` to `0x7E`. Octets with all bits set to zero MAY follow the string value when needed, such as reducing the length of a stored string while maintaining the same `Element Data Size`. A string with one or more octets with all bits set to zero and a string without one or more octets with all bits set to zero are semantically equal.
+A `String Element` MUST either be empty (zero-length) or contain printable ASCII characters [@!RFC0020] in the range of `0x20` to `0x7E`, with an exception made for termination (see [the section on the `Terminating Elements`](#terminating-elements)).
 
 ## UTF-8 Element
 
 A `UTF-8 Element` MUST declare a length in octets from zero to `VINTMAX`. If the `EBML Element` is not defined to have a `default` value, then a `UTF-8 Element` with a zero-octet length represents an empty string.
 
-A `UTF-8 Element` contains only a valid Unicode string as defined in [@!RFC3629]. Octets with all bits set to zero MAY follow the string value when needed, such as reducing the length of a stored UTF-8 data while maintaining the same `Element Data Size`. A UTF-8 value with one or more octets with all bits set to zero and a UTF-8 value without one or more octets with all bits set to zero are semantically equal.
+A `UTF-8 Element` contains only a valid Unicode string as defined in [@!RFC3629], with an exception made for termination (see [the section on the `Terminating Elements`](#terminating-elements)).
 
 ## Date Element
 
@@ -248,6 +248,19 @@ The `Master Element` contains zero, one, or many other elements. `EBML Elements`
 A `Binary Element` MUST declare a length in octets from zero to `VINTMAX`.
 
 The contents of a `Binary Element` should not be interpreted by the `EBML Reader`.
+
+# Terminating Elements
+
+`Null Octets`, which are octets with all bits set to zero, MAY follow the value of a `String Element` or `UTF-8 Element` to serve as a terminator. An `EBML Writer` MAY terminate a `String Element` or `UTF-8 Element` with `Null Octets` in order to overwrite a stored value with a new value of lesser length while maintaining the same `Element Data Size` (this can prevent the need to rewrite large portions of an `EBML Document`); otherwise the use of `Null Octets` within a `String Element` or `UTF-8 Element` is NOT RECOMMENDED. An `EBML Reader` MUST consider the value of the `String Element` or `UTF-8 Element` to be terminated upon the first read `Null Octet` and MUST ignore any data following the first `Null Octet` within that `Element`. A string value and a copy of that string value terminated by one or more `Null Octets` are semantically equal.
+
+The following table shows examples of semantics and validation for the use of `Null Octets`. Values to represent `Stored Values` and the `Semantic Meaning` as represented as hexidecimal values.
+
+Stored Value        | Semantic Meaning
+:-------------------|:-------------------
+0x65 0x62 0x6d 0x6c | 0x65 0x62 0x6d 0x6c
+0x65 0x62 0x00 0x6c | 0x65 0x62
+0x65 0x62 0x00 0x00 | 0x65 0x62
+0x65 0x62           | 0x65 0x62
 
 # EBML Document
 
--- a/specification.markdown
+++ b/specification.markdown
@@ -262,6 +262,55 @@ Stored Value        | Semantic Meaning
 0x65 0x62 0x00 0x00 | 0x65 0x62
 0x65 0x62           | 0x65 0x62
 
+# Guidelines for Updating Elements
+
+An EBML Document can be updated without requiring that the entire EBML Document be rewritten. These recommendations describe strategies to change the `Element Data` of a written `EBML Element` with minimal disruption to the rest of the `EBML Document`.
+
+## Reducing a Element Data in Size
+
+There are three methods to reduce the size of `Element Data` of a written `EBML Element`.
+
+### Adding a Void Element
+
+When an `EBML Element` is changed to reduce its total length by more than one octet, an `EBML Writer` SHOULD fill the freed space with a `Void Element`.
+
+### Extending the Element Data Size
+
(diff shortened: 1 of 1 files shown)


dericed:
ping

specification.markdown
```diff
@@ -245,6 +245,18 @@ A `Binary Element` MUST declare a length in octets from zero to `VINTMAX`.
 
 The contents of a `Binary Element` should not be interpreted by the `EBML Reader`.
 
+# Terminating Elements
+
+`Null Octets`, which are octets with all bits set to zero, MAY follow the value of a `String Element` or `UTF-8 Element` to serve as a terminator. An `EBML Writer` MAY terminate a `String Element` or `UTF-8 Element` with `Null Octets` in order to overwrite a stored value with a new value of lesser length while maintaining the same `Element Data Size` (this can prevent the need to rewrite large portions of an `EBML Document`); otherwise the use of `Null Octets` within a `String Element` or `UTF-8 Element` is NOT RECOMMENDED. An `EBML Reader` MUST consider the value of the `String Element` or `UTF-8 Element` to be terminated upon the first read `Null Octet` and ignore any following data within that `Element`. If an `EBML Writer` terminates the value of a `String Element` or `UTF-8 Element` with a `Null Octet` then any subsequent octets of the `Element` MUST also be `Null Octets`. A string value and a copy of that string value terminated by one or more `Null Octets` are semantically equal.
```
mbunkus: About "this can prevent the need to rewrite large portions of an `EBML Document`" and the whole sentence. I don't like that part because it promotes the wrong method of overwriting elements. We already have a perfectly valid element for such a use case: `EbmlVoid`. As the `ElementID` of `EbmlVoid` is only a single byte, the shortest-possible ID, you can use an `EbmlVoid` to overwrite any element in any situation without having to re-write or overwrite anything else.

I'm OK with allowing such termination, but I'd rather we don't promote it. I'd vote for ending the sentence much ealier, e.g.:

> An `EBML Writer` MAY terminate a `String Element` or `UTF-8 Element` with `Null Octets`, but it is NOT RECOMMENDED.

We could even state that instead of terminating an existing element in order to make it empty one should use an `EbmlVoid`.

mbunkus: > If an `EBML Writer` terminates the value of a `String Element` or `UTF-8 Element` with a `Null Octet` then any subsequent octets of the `Element` MUST also be `Null Octets`.

Missing comma before "then". However, I don't think this is the right thing to do. If we make it a `MUST`, then a reader has to consider a file with e.g. `0x00 0x4a` invalid. I known programmers (I'm one of them). We're lazy. I'm sure there are programs that simply overwrite the first byte of a string with `0x00` and leave the rest as-is, causing such a file to become invalid with the proposed verbiage.

I'd rather say we treat such content the same as C would treat such a string: all bytes starting with the `0x00` byte must be ignored, no matter their content.

As a consequence the following table's "Validity" column should be removed.

dericed: I think I'll add a new section on recommendation order for handling element overwrites. If the octet length is decreased by >1 octets, then the void can be used. But if the octet length is reduced by 1 octet, then what is preference order:

- add null octet to end
- reduce vint efficiency of the element data size (so element data size uses +1 octets and then the element data is shifted accordingly)
- other methods?

mbunkus: What mkvpropedit and the GUI's header/chapter editors do in a case of having to shrink by one byte is to move the following element's header (ID + size field) up one byte and to extend its size field by one byte. This works because both programs only rewrite level 1 elements that way; therefore they don't have to care about having to adjust the upper level, too.

The problem with writing a null value at the end is that we've considered that to be invalid if I remember correctly as that byte wouldn't be a valid start of an element ID.

Re-writing the element to shrink and extending its size field by one byte seems to the most reasonable regarding a general description.

I wouldn't object to allowing null bytes between elements and requiring readers to simply skip such null bytes.

dericed: Re: "The problem with writing a null value at the end is that we've considered that to be invalid if I remember correctly as that byte wouldn't be a valid start of an element ID."

If a string element was reduced by one octet the last octet would be null, but the Element Data Size of that element would be unchanged since it includes the null octet. I wasn't suggesting allowing null bytes between elements. In the current EBML draft we have

> Element Data stored within Master Elements SHOULD only consist of EBML Elements and SHOULD NOT contain any data that is not part of an EBML Element. When EBML is used in transmission or streaming, data that is not part of an EBML Element is permitted to be present within a Master Element if unknownsizeallowed is enabled within the definition for that Master Element.

so the only places where null octets are allowed `in between` elements would be in ones that support unknown sizes (i.e. Segment and Cluster) and no others.

mbunkus: That's what I was thinking about, yeah.

I tried to answer the general question of what our recommendation should be if any element (not necessarily one of the two string types) needs to be extended/shrunk. It can easily happen for master elements.

This leaves us with only one viable strategy that I can see at the moment: re-writing the element located before that one byte hole and extending its vint size.

dericed:
To clarify. If for example the `CodecInfoURL` contains `ietf.org/` and the intent is to edit to `ietf.org` (remove trailing slash) then:

To reduce by 1 octet


| Element ID | Element Data Size | Element Data       | Trailing Junk | Status                                              |
|------------|-------------------|--------------------|---------------|-----------------------------------------------------|
| 0x3B4040   | 0x88              | 0x696574662e6f7267 | 00            | Invalid in Elements that do not permit unknown size, not recommended otherwise |
| 0x3B4040   | 0x88              | 0x696574662e6f7200 |               | Valid by not recommended |
| 0x3B4040   | 0x4007            | 0x696574662e6f72   |               | Valid, recommended       |

mbunkus:
That looks reasonable to me. We should mention that the second method is only valid for the two string types.

dericed:
I added a section about reducing Element Data in length during file edits, so that alternatives to Terminating Strings are sufficiently hyped.

mbunkus:
I'm fine with the whole section "Guidelines for Updating Elements".

My outstanding concern is the question whether string elements with null bytes in the middle are invalid. As no one has said anything about this concern yet, I'll reiterate what I've written:

> If an `EBML Writer` terminates the value of a `String Element` or `UTF-8 Element` with a `Null Octet` then any subsequent octets of the `Element` MUST also be `Null Octets`.

I don't think this is the right thing to do. If we make it a MUST, then a reader has to consider a file with e.g. 0x00 0x4a invalid. I known programmers (I'm one of them). We're lazy. I'm sure there are programs that simply overwrite the first byte of a string with 0x00 and leave the rest as-is, causing such a file to become invalid with the proposed verbiage.

I'd rather say we treat such content the same as C would treat such a string: all bytes starting with the 0x00 byte must be ignored, no matter their content.

As a consequence the following table's "Validity" column should be removed.

retokromer:
I guess, in the table at https://github.com/Matroska-Org/ebml-specification/pull/141#issuecomment-301495416 it should read `Valid but not recommended`.

robUx4:
I agree with https://github.com/Matroska-Org/ebml-specification/pull/141#issuecomment-302073954 This section is probably overkill and unnecessary with:
```An `EBML Reader` MUST consider the value of the `String Element` or `UTF-8 Element` to be terminated upon the first read `Null Octet` and ignore any following data within that `Element`.```

**Comment by dericed** [comment]
I'm not sure how to proceed based on the comments. @robUx4 which part is overkill? Should I break the PR up into smaller ones?

**Comment by robUx4** [comment]
`any subsequent octets of the Element MUST also be Null Octets` That part can be removed IMO.
And like @mbunkus said, it's probably to reword it in a way that anything after the first 0x00 must be ignored.

**Comment by dericed** [comment]
I removed the sentence containing `any subsequent octets of the Element MUST also be Null Octets` locally and adjusted an existing sentence in the patch to say

> An `EBML Reader` MUST consider the value of the `String Element` or `UTF-8 Element` to be terminated upon the first read `Null Octet` and MUST ignore any data following the first `Null Octet` within that `Element`.

Because of this change (clarification that non-null octets after null octets are okay but should be ignored) changes the following table since now all entries in it are valid. Thus I'm removing the validity column there. Patch is updated for a new review. Any comments on the new `# Guidelines for Updating Elements` section?

File: specification.markdown [hunk 2]
```diff
@@ -245,6 +245,68 @@ A `Binary Element` MUST declare a length in octets from zero to `VINTMAX`.
...
+In the following table, a four octet long `Element Data` is changed to a three octet long value followed by a `Null Octet`; the `Element Data Size` includes any `Null Octets` used to terminate `Element Data` so remains unchanged.
+
+| Status      | Element ID | Element Data Size | Element Data       |
+|-------------|------------|-------------------|--------------------|
+| Before edit | 0x3B4040   | 0x84              | 0x65626d6c         |
+| After edit  | 0x3B4040   | 0x84              | 0x6d6b7600         |
+
+Note that this method is NOT RECOMMENDED. For reductions of one octet, the method for `Extending the Element Data Size` SHOULD be used. For reduction by more than one octet, the method for `Adding a Void Element` SHOULD be used.
+
+## Considerations of Fixity when Updating Elements
```
robUx4: Is `Fixity` a term that should be defined somewhere ? I'm not sure most people know what it means.

robUx4:
Apart from Fixity (not blocking) LGTM.

dericed:
Good point. Updated to `Considerations when Updating Elements with CRC`.

@mbunkus, any comments on the updates?

dericed:
I updated this PR to resolve a conflict. Please re-review. Nudge to @mbunkus

mbunkus:
I have no further objections. The changes look good to me.

dericed:
Thanks, merging then. Btw Monday is the deadline for a new version of this document for consideration at IETF99.

Issue #148, opened 2017-05-28
PDF of version 02


retokromer:
The page 23 is rendered as… miniature.

PR #149, opened 2017-06-02
Line wrapping fixes

dericed:
breaks alignment but fixes compliance with chars-per-line limit

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -320,7 +320,7 @@ The path defines the allowed storage locations of the `EBML Element` within an `
 The `path` attribute is REQUIRED.
 
 ```
-EBMLFullPath             = EBMLElementOccurrence "(" EBMLReferencePath ")"
+EBMLFullPath           = EBMLElementOccurrence "(" EBMLReferencePath ")"
 EBMLReferencePath        = [EBMLParentPath] EBMLElementPath
 EBMLParentPath           = EBMLFixedParent EBMLLastParent
 EBMLFixedParent          = *(EBMLPathAtom)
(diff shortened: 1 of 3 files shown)


specification.markdown
```diff
@@ -320,7 +320,7 @@ The path defines the allowed storage locations of the `EBML Element` within an `
 The `path` attribute is REQUIRED.
 
 ```
-EBMLFullPath             = EBMLElementOccurrence "(" EBMLReferencePath ")"
+EBMLFullPath           = EBMLElementOccurrence "(" EBMLReferencePath ")"
```
retokromer: LUTM, but fixes the issue (U = ugly ;-)

retokromer:
fixes https://github.com/Matroska-Org/ebml-specification/issues/148

specification.markdown
dericed: i can't think of a better idea, unless @robUx4 wants to rename one of these components to use a shorter string length.

robUx4: Unless @retokromer comes with something less U, I think it's fine.

robUx4:
LGTM although it might break the ugly code that's generating various code from the XML source.

specification.markdown
dericed: Could also resolve by renaming `VariableParentOccurrence` to `VarParentOccurrence`.

retokromer: No, I don’t have a proposal having less U. As Dave said, only shorter names would prevent it. That said, this fixes the issue of the «miniature pages».

dericed: I think shorter names are less `U` than breaking alignment.

Issue #128, opened 2017-01-08
how to constrain EBML Header with an EBML Schema


dericed:
This issue is based on an initial discussion at https://github.com/Matroska-Org/ebml-specification/issues/98#issuecomment-249088418.

The `EBML Schema` section claims:

> The `EBML Schema` does not itself document the `EBML Header`, but documents all data of the `EBML Document` that follows the `EBML Header`. The `EBML Header` itself is documented by this specification in the `EBML Header Elements` (see [EBML Header Elements](#ebml-header-elements)). The `EBML Schema` also does not document `Global Elements` that are defined by this document (namely the `Void Element` and the `CRC-32 Element`).

However we have cases where EBML Document Types (Matroska) add additional constraints to the EBML Header. For example, EBML allows EBMLMaxSizeLength=9 but Matroska doesn't.

Should we change the above noted paragraph to say:

> The `EBML Schema` MUST begin with a definition of the `EBML Header` and EBML's `Global Elements` which is followed with a definition of the `EBML Body`. The `EBML Schema` MUST document the ` of the `EBML Header` as documented by this specification in the `EBML Header Elements` (see [EBML Header Elements](#ebml-header-elements)), but MAY constrain those `Elements` by adding or constraining the `range` attributes. For example, an `EBML Schema` MAY constrain the `EBMLMaxSizeLength` to a maximum value of `8`.

The example schema would then need to be updated to include the Header and Global Elements. The downside is that this causes some redundancy between the EBML spec and the EBML Schema, but it makes the EBML Schema less dependent on the EBML spec. Also the old specdata.xml documented the EBML Header and Global Elements in addition to the Body.

dericed:
ping @robUx4, opinion about this change?

dericed:
ping to @robUx4 again. In your view should we say that all EBML Schema must include the EBML Header elements?

robUx4:
`MUST document the of the EBML Header` there's a typo in there.
I agree with the general goal and implementation.

retokromer:
I guess, it’s fine.

robUx4:
As mentioned in https://github.com/Matroska-Org/matroska-specification/pull/135#issuecomment-307703820, do we want to force writing the whole EBML header definition for each format or only the differing values ?

dericed:
Good point. OK I'll update the patch to reflect only storing the differences, i.e. constraints.

dericed:
[...]

PR #152, opened 2017-06-11
require definition of EBML Header/Globals in EBML Schema

dericed:
related to https://github.com/Matroska-Org/ebml-specification/issues/128

### DIFF
--- a/ebml_schema_example.xml
+++ b/ebml_schema_example.xml
@@ -1,6 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <EBMLSchema xmlns="https://ietf.org/cellar/ebml"
   docType="files-in-ebml-demo" version="1">
+ <!-- constraints to the range of two EBML Header Elements -->
+ <element name="EBMLReadVersion" path="1*1(\EBML\EBMLReadVersion)"
+   id="0x42F7" minOccurs="1" maxOccurs="1" range="1" default="1"
+   type="uinteger"/>
+ <element name="EBMLMaxSizeLength" path="1*1(\EBML\EBMLMaxSizeLength)"
+   id="0x42F3" minOccurs="1" maxOccurs="1" range="8" default="8"
+   type="uinteger"/>
  <!-- Root Element-->
  <element name="Files" path="*1(\Files)" id="0x1946696C" type="master">
   <documentation lang="en" type="definition">Container of data and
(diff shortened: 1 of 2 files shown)


ebml_schema_example.xml
```diff
@@ -1,6 +1,31 @@
...
+ <element name="EBMLMaxSizeLength" path="1*1(\EBML\EBMLMaxSizeLength)"
+   id="0x42F3" minOccurs="1" maxOccurs="1" range="not 0" default="8"
+   type="uinteger"/>
+ <element name="DocType" path="1*1(\EBML\DocType)" id="0x4282"
+   minOccurs="1" maxOccurs="1" size=">0" type="string"/>
+ <element name="DocTypeVersion" path="1*1(\EBML\DocTypeVersion)"
+   id="0x4287" minOccurs="1" maxOccurs="1" default="1" type="uinteger"/>
+ <element name="DocTypeReadVersion" path="1*1(\EBML\DocTypeReadVersion)"
+   id="0x4285" minOccurs="1" maxOccurs="1" default="1" type="uinteger"/>
+ <element name="CRC-32" path="1*1(\EBML\EBMLReadVersion)" id="0xBF"
```
robUx4: I think the path for the CRC-32 is wrong

dericed:
Oops, may be removing the line altogether since there's no Matroska constaint on EBML's CRC. Will ping once updated.

dericed:
This PR is updated based upon @robUx4's comments at https://github.com/Matroska-Org/matroska-specification/pull/135#issuecomment-307703820. Now the documentation only mandates EBML Header elements are defined in an EBML Schema if there is an added constraint to the use of that element. Also only constraints to the range attribute are allowed. Thus the matroska EBML Schema can constrain the range of EBMLMaxSizeLength to 8, but doesn't need to document all the other Header Elements as well. I left out constraints to EBML's global elements since it doesn't really make sense to constrain them.

specification.markdown
```diff
@@ -279,7 +279,9 @@ An `EBML Schema` is an XML Document that defines the properties, arrangement, an
...
+The `EBML Schema` MUST document all Elements of the `EBML Body`. The `EBML Schema` does not document `Global Elements` that are defined by this document (namely the `Void Element` and the `CRC-32 Element`).
+
+An `EBML Schema` MAY constrain the use of `EBML Header Elements` (see [EBML Header Elements](#ebml-header-elements)) by adding or constraining that Element's `range` attribute. For example, an `EBML Schema` MAY constrain the `EBMLMaxSizeLength` to a maximum value of `8` or MAY constain the `EBMLVersion` to only support a value of `1`. If an `EBML Schema` adopts the `EBML Header Element` as-is, then it is not REQUIRED to document that Element within the `EBML Schema`. If an `EBML Schema` constrains the range of an `EBML Header Element` then that `Element` MUST be documented within an `<element>` node of the `EBML Schema`. This document provides an example of an `EBML Schema`, see [EBML Schema Example](#ebml-schema-example).
```
mbunkus: A missing comma before "then that `Element` MUST be documented…"

dericed:
Updated per @mbunkus' comment at https://github.com/Matroska-Org/ebml-specification/pull/152#discussion_r125188430.

PR #154, opened 2017-07-02
bump to version 03

dericed:
I suggest merging this before Monday so that we can upload a new
version of the draft to the document tracker before the doc freeze for
IETF99.
If possible let's review/merge https://github.com/Matroska-Org/ebml-specification/pull/152 first.


retokromer:
I agree.

PR #155, opened 2017-07-02
Clarify the Header structure

lu-zero:
Try to rephrase so it is clear that the Header contains not only a single Master Element and only that.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -321,7 +321,7 @@ The `EBML Header` is a declaration that provides processing instructions and ide
 
 The `EBML Header` documents the `EBML Schema` (also known as the `EBML DocType`) that is used to semantically interpret the structure and meaning of the `EBML Document`. Additionally the `EBML Header` documents the versions of both `EBML` and the `EBML Schema` that were used to write the `EBML Document` and the versions required to read the `EBML Document`.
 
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -321,7 +321,7 @@ The `EBML Header` is a declaration that provides processing instructions and ide
 
 The `EBML Header` documents the `EBML Schema` (also known as the `EBML DocType`) that is used to semantically interpret the structure and meaning of the `EBML Document`. Additionally the `EBML Header` documents the versions of both `EBML` and the `EBML Schema` that were used to write the `EBML Document` and the versions required to read the `EBML Document`.
 
-The `EBML Header` consists of a single `Master Element` with an `Element Name` of `EBML` and `Element ID` of `0x1A45DFA3` (see [the definition of the `EBML` Element](#ebml-element)). The `EBML Header` MUST only contain `EBML Elements` that are defined as part of this document.
+The `EBML Header` MUST contain a single `Master Element` with an `Element Name` of `EBML` and `Element ID` of `0x1A45DFA3` (see [the definition of the `EBML` Element](#ebml-element)) and any number of additional `EBML Elements`. The `EBML Header` MUST only contain `EBML Elements` that are defined as part of this document.
```
retokromer: `MUST contain` […] `any number` sounds strange to me (but I am not English native).

lu-zero:
@mbunkus you are right, I shifted the ambiguity by 1 paragraph.

PR #163, opened 2017-08-08
adds void element to defined terms list

### LINKED ISSUE #161: Add `Void Element` to defined terms?
In the opening Notations and Conventions section, I think the `Void Element` warrants an entry. It is referenced within the term-definition list (under Root Element) and in the security section below, then referenced at the EBML Schema level (with appropriate link to its definition). It seems substantial enough to the specification to be noted at the top with all of the others.

ablwr:
With bonus whitespace elimination?

Resolves #161

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -60,6 +60,8 @@ This document defines specific terms in order to define the format and applicati
 
 `Descendant Element`: A `Descendant Element` is a relative term to describe any `EBML Elements` contained within a `Master Element`, including any of the `Child Elements` of its `Child Elements`, and so on.
 
+`Void Element`: A `Void Element` is an `Element` used to overwrite damaged data or reserve space within a `Master Element` for later use.
+
 `Element Name`: The official human-readable name of the `EBML Element`.
 
 `Element Path`: The hierarchy of `Parent Element` where the `EBML Element` is expected to be found in the `EBML Body`.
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -60,6 +60,8 @@ This document defines specific terms in order to define the format and applicati
 
 `Descendant Element`: A `Descendant Element` is a relative term to describe any `EBML Elements` contained within a `Master Element`, including any of the `Child Elements` of its `Child Elements`, and so on.
 
+`Void Element`: A `Void Element` is an element used to void damaged data or reserve space in a sub-element for later use.
```
dericed: Maybe better to avoid the `void` in the definition, perhaps overwrite? Also element->`Element`. The term sub-element may be misused. I'd think this is like a `Child Element` but that doesn't make sense. Maybe you mean within a `Master Element`. Such as `<Segment><Void/><Tracks/><Clusters/></Segment>`.

ablwr: I looked through the rest of the spec to see what I could come up with on my own, just saw that it can be but is not considered officially to be a `Root Element`, so seems like it isn't necessarily a `Child` either?

ablwr: Ah gotcha, re: Master element. Updating now.

ablwr:
@mbunkus may want to re-review

ablwr:
Mysteries of the `Void`...

PR #166, opened 2017-09-03
expand on Global Element section

dericed:


specification.markdown
```diff
@@ -743,7 +743,9 @@ type: Unsigned Integer
 
 description: The minimum `DocType` version an `EBML Reader` has to support to read this `EBML Document`. The value of the `DocTypeReadVersion Element` MUST be less than or equal to the value of the `DocTypeVersion Element`.
 
-## Global elements (used everywhere in the format)
+## Global Elements
+
+EBML defines these `Global Elements` which MAY be stored within any Element of an `EBML Document` as defined by their `Element Path`.
```
mbunkus: Maybe rather say "within any Master Element"? Just to be extra precise.

mbunkus (dismissed):

ablwr:
If global elements are only possible, should you use language stronger than MAY? MAY implies "truly optional."

robUx4:
The `MAY` refers to the `be stored` not the fact that it MUST be in an `Master Element`, if that's what you mean.

ablwr:
Ahhh, of course, my mistake. It's still early here. 😅

mbunkus (approved):
LGTM with that commit from Steve.

PR #168, opened 2017-11-28
gives grammatical consistency to definitions list in Section 2

ablwr:
This has come up before, maybe in matroska-specification, in how to handle definitions. This PR updates the definitions so they all follow the same pattern and do not repeat the word being defined within the definition.


ablwr:
this also removes a trailing space, sorry about that. My editor has it turned on by default, and I turned it off for now (although after this is merged, there won't be any trailing whitespace to get rid of).

PR #171, opened 2017-11-29
adds header for crc-32 element, standardizes global elemement header

ablwr:
CRC-32's header was missing, and I simplified the global header above it to match the EBML Header Elements H2 level heading above.


ablwr:
Oh! This was already done, I must have been working on a stale version! Closing!

PR #174, opened 2018-01-25
add a page with ideas for the future development

retokromer:


dericed:
I think having this in an issue as opposed to a commit may offer advantages since we could associate it with issue discussion, labels, milestones, status, etc.

retokromer:
The experiences made were very bad, but let’s give a try https://github.com/Matroska-Org/ebml-specification/issues/175.

Issue #48, opened 2016-01-10
add section for handling bad data


dericed:
For instance what to do when finding a bad CRC value or a duplicate Element that isn't allowed to appear multiple times.

dericed:
nudging on this as it's worth offering guidance on these scenarios. If reading a master element which contains a CRC-32 element which stores an invalid value, what should be reader do: ignore the master element, reject the file, read it anyhow?
If a 2nd occurrence of a maxoccurs=1 element is found, which should take precedence?

retokromer:
One possible answer from the real world (our own reading tools):

- give an error message having the options: try to read anyway or abort;
- if additional 2nd, then give an error message and simply ignore 2nd.

robUx4:
If a master element has a CRC-32, sub elements may have some too. If these are valid the data for these can be used. The rest of the master data should no be used. It's not that big a deal to handle since you only know if the CRC-32 matches when you have read all the data. So you can tell which parts are OK and which are not.
Not easy to phrase properly though.

PR #177, opened 2018-02-12
start section on reading bad data

dericed:
addresses https://github.com/Matroska-Org/ebml-specification/issues/48

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -787,7 +787,7 @@ If a `Master Element` contains a `CRC-32 Element` that doesn't validate, then th
 
 If a `Master Element` contains more occurrences of a `Child Master Element` than permitted according to the `maxOccurs` and `recurring` attributes of the definition of that `Element` then the occurrences in addition to `maxOccurs` MAY be ignored.
 
-If a `Master Element` contains more occurrences of a `Child Element` that is not a `Master Element` than permitted according to the `maxOccurs` attribute of the definition of that `Element` then the instances of that `Element` with the greatest byte offset from the beginning of its `Parent Element` SHOULD be ignored.
+If a `Master Element` contains more occurrences of a `Child Element` that is not a `Master Element` than permitted according to the `maxOccurs` attribute of the definition of that `Element` then all but the instance of that `Element` with the smallest byte offset from the beginning of its `Parent Element` SHOULD be ignored.
 
 # IANA Considerations
 
--- a/specification.markdown
+++ b/specification.markdown
@@ -779,6 +779,16 @@ type: Binary
 
 description: Used to void damaged data, to avoid unexpected behaviors when using damaged data. The content is discarded. Also used to reserve space in a sub-element for later use.
 
+# Considerations for Reading EBML Data
+
+The following scenarios describe events to consider when reading `EBML Documents` and the recommended design of an `EBML Reader`.
+
+If a `Master Element` contains a `CRC-32 Element` that doesn't validate, then the `EBML Reader` MAY ignore all contained data except for `Descendant Elements` which contain their own valid `CRC-32 Element`.
+
+If a `Master Element` contains more occurrences of a `Child Master Element` than permitted according to the `maxOccurs` and `recurring` attributes of the definition of that `Element` then the occurrences in addition to `maxOccurs` MAY be ignored.
+
+If a `Master Element` contains more occurrences of a `Child Element` that is not a `Master Element` than permitted according to the `maxOccurs` attribute of the definition of that `Element` then the instances of that `Element` with the greatest byte offset from the beginning of its `Parent Element` SHOULD be ignored.
+
 # IANA Considerations
 
 This document creates a new IANA Registry called "CELLAR EBML Element ID Registry".


specification.markdown
```diff
@@ -778,3 +778,14 @@ minOccurs: 0
 type: Binary
 
 description: Used to void damaged data, to avoid unexpected behaviors when using damaged data. The content is discarded. Also used to reserve space in a sub-element for later use.
+
+# Considerations for Reading EBML Data
+
+The following scenarios describe events when reading `EBML Documents` and the REQUIRED behavior by an `EBML Reader`.
+
+If a `Master Element` contains a `CRC-32 Element` that doesn't validate, then the `EBML Reader` MUST ignore all contained data except for `Descendant Elements` which contain their own valid `CRC-32 Element`.
```
robUx4: That's only for systems that don't want to handle damaged data. A player (let's call it VLC) will try to read the data anyway and extract as much as it finds valid. You cannot say that players should ignore a Cluster because of single byte (P2P unfinished download for example) is missing.

specification.markdown
```diff
@@ -778,3 +778,14 @@ minOccurs: 0
...
+
+If a `Master Element` contains more occurrences of a `Child Master Element` than permitted according to the `maxOccurs` and `recurring` attributes of the definition of that `Element` then the occurrences in addition to `maxOccurs` MUST be ignored.
```
robUx4: `SHOULD` ? Although it would create different behaviors depending on the internal limit they set.

specification.markdown
```diff
@@ -778,3 +778,14 @@ minOccurs: 0
...
+
+If a `Master Element` contains more occurrences of a `Child Element` that is not a `Master Element` than permitted according to the `maxOccurs` attribute of the definition of that `Element` then occurrences in addition to `maxOccurs` MUST be ignored.
```
robUx4: Not sure. So far the consensus was to use the last value, ie overwrite the previous one. That's as long as maxOccurs is 1.
`SeekHead` has maxOccurs to `2` but since it's a level 1 element it's treated differently...

mbunkus:
I don't have any strong feelings for any of the possible ways to treat excessive elements, to be honest. Either way would be fine with me.

I do agree that `MUST` is to strong regarding masters with invalid CRCs. Let reading applications scrape as much as possible; that's usually what users want.

mbunkus:
I've just re-read this PR. In my opinion…

* excessive elements MUST be ignored, both for Masters and non Masters (because this is the easiest way to reason about how a reader has to act, especially regarding Masters and the question how to handle different content in the Masters — e.g. a sub-Master is present in the first Master but not in the second Master which is semantically different)
* Masters with broken CRC SHOULD be ignored/dropped (to allow recovery of as much data as possible for readers that support handling broken files).

dericed:
If I'm reading correctly there is some conflict between the suggestions of @robUx4 and @mbunkus.

Regarding handling of repetitions of an element within a parent when maxoccurs=1.
@mbunkus "excessive elements MUST be ignored..."
@robUx4 " So far the consensus was to use the last value, ie overwrite the previous one."

Regarding handling master elements with broken CRC:
@mbunkus "Masters with broken CRC SHOULD be ignored/dropped..."
@robUx4 "That's only for systems that don't want to handle damaged data. A player (let's call it VLC) will try to read the data anyway and extract as much as it finds valid. You cannot say that players should ignore a Cluster because of single byte (P2P unfinished download for example) is missing."

Suggestions for consensus on a recommendation here?

mbunkus:
Regarding the CRC: I think Steve & I are pretty much in agreement here. I'm not strictly in favor of using "…SHOULD…" here — my main issue is with the PR's proposal of using "MUST" which I consider way too strong.

We could split up the recommendation to handle to different reader systems: one that tries to handle as much data as possible and that's prepared to handle potentially broken data, and one that tries to be conservative, err on the safe side.

For the former type of system (the one designed to handle broken data) I don't think the specs actually need to mandate a way to handle things. Just let the system deal with it as best as it can.

For the latter type, the "play it safe" type, we might keep the "Master elements with incorrect CRCs MUST be dropped" language.

I'm not totally opposed to "MUST". I'm only opposed to mandate "MUST" for each situation. And that's how I understood Steve, too.

As for the maxoccurs: I think it's silly to mandate that later copies of an element take precedence over earlier ones as that would force reader systems to actively look for them instead of being able to use the elements as soon as they're encountered. On the other hand, saying that any further copy MUST be ignored would force a reader system to keep track of which elements it has already encountered and which it hasn't seen yet. I don't know of a single implementation that acts this way: mkvmerge, ffmpeg, VLC all use the last element found. Declaring what they're doing wrong and themselves to be out of compliance might be received badly.

Maybe a compromise might be something like "undefined behavior" in programming languages. The fact that excessive instances exist makes the file invalid already; so just let the implementation decide what to do with that.

dericed:
I forced an update on this pr to resolve conflicts. I also adjusted the language more towards recommendations rather than requirements. Could use a re-review.

dericed:
I don't have a strong preference for preferred the first or the last instance of an element which otherwise shouldn't recur, but I think it would be helpful to express a preference to a reader who may be developing an EBML Reader.

robUx4:
> Regarding the CRC: I think Steve & I are pretty much in agreement here. I'm not strictly in favor of using "…SHOULD…" here — my main issue is with the PR's proposal of using "MUST" which I consider way too strong.

I agree, `SHOULD` is the proper "strength" for the CRC. `MAY` could also work. But let's keep SHOULD.

robUx4:
(OK I see you use MAY, so it's fine for me)

specification.markdown
```diff
@@ -779,6 +779,16 @@ type: Binary
...
 
+# Considerations for Reading EBML Data
+
+The following scenarios describe events to consider when reading `EBML Documents` and the recommended design of an `EBML Reader`.
+
+If a `Master Element` contains a `CRC-32 Element` that doesn't validate, then the `EBML Reader` MAY ignore all contained data except for `Descendant Elements` which contain their own valid `CRC-32 Element`.
+
+If a `Master Element` contains more occurrences of a `Child Master Element` than permitted according to the `maxOccurs` and `recurring` attributes of the definition of that `Element` then the occurrences in addition to `maxOccurs` MAY be ignored.
+
+If a `Master Element` contains more occurrences of a `Child Element` that is not a `Master Element` than permitted according to the `maxOccurs` attribute of the definition of that `Element` then the instance of that `Element` with the greatest byte offset from the beginning of its `Parent Element` SHALL be given preference to supersede prior instances of that `Element`.
```
robUx4: > that is not a `Master Element`

Not too sure about that. Why do we need this exception ?

mbunkus: How surplus `Child Master Elements` are handled is already explained in the paragraph above.

mbunkus:
The reason I prefer to take the first instead of the last element is simple: simplicity and latency, as in "the reader can act on an element as soon as it finds it, it doesn't have to look for further occurrences of it".

Let's take the cluster timestamp as an example. It's specified as maxOccurs=1. If we say that the last SHOULD be used instead of the first, a reader following the EBML specs/recommendations would have to scan the whole cluster before starting to play the first frame of the cluster, which could take several seconds, especially in live streaming situations.

Or to put it differently: as soon as it finds an element with maxOccurs=1, there are only three possible further outcomes:

1. The element does not occur again. The reader can use the first occurrence's value.
2. The element does occur again with the same value as the first one. The file is invalid (as the second occurrence violates maxOccurs=1). It doesn't matter in this case: the reader can use the first occurrence's value.
3. The element does occur again with a different value than the first one. The file is even more invalid (as the second occurrence violates maxOccurs=1 and their values differ). In this case there's no clear right thing to do, and the reader can use the first occurrence's value.

In all cases simply using the value the first time it's seen is the sensible thing to do. Recommending to use the last one doesn't improve things one bit.

robUx4:
So you would replace
> SHALL be given preference to supersede prior instances of that `Element`.

by

> SHOULD be ignored.

?

Also instance*s* should be plural since `maxOccurs` may be bigger than 1.

dericed:
That sounds fine with me, but does that conflict with @mbunkus earlier saying
> On the other hand, saying that any further copy MUST be ignored would force a reader system to keep track of which elements it has already encountered and which it hasn't seen yet. I don't know of a single implementation that acts this way: mkvmerge, ffmpeg, VLC all use the last element found.

is changing MUST to SHOULD enough to resolve this concern?

mbunkus:
> So you would replace
>
>> SHALL be given preference to supersede prior instances of that Element.
>
> by
>
>> SHOULD be ignored.
>
> ?

Yes, I'd prefer this wording.

> Also instances should be plural since maxOccurs may be bigger than 1.

Indeed.

> is changing MUST to SHOULD enough to resolve this concern?

For me it is. A SHOULD can be ignored if there are good reasons; simplicity of implementation (not having to track which elements have been seen yet) might be one of those reasons. MAY would be too weak a recommendation in my opinion.

robUx4:
IMO SHOULD means that it's OK not to do it if you want. We're talking about cases that violates the rule anyway. We can't really have rules outside of the rules themselves.

dericed:
PR is updated according to most recent comments.

mbunkus:
> We can't really have rules outside of the rules themselves.

Exactly.

> PR is updated according to most recent comments.

Thanks, however…

> …then the instances of that `Element` with the greatest byte offset from the beginning of its `Parent Element` SHOULD be ignored.

Only the first one should be used, meaning:  "…then all but the instance of that `Element` with the smallest byte offset from the beginning of its `Parent Element` SHOULD be ignored."

Otherwise you'd have strange situations such as having four instances of an element, the specs say only the last one SHOULD be ignored, and you'd wonder which of the first three instances to use.

dericed:
@mbunkus Thanks for catching that. I added a separate commit to fix that with your wording. For whoever merges, please squash when merging.

dericed:
ping to @robUx4 for re-approval

robUx4:
pong
I'm reading

robUx4:
> all but

Makes it hard to read. But I can't find a cleaner way to do it. So I'm OK.

Issue #157, opened 2017-07-20
Define DoctypeVersion 0 as invalid/experimental


robUx4:
This was defined first as a Matroska issue but it seems to be more for EBML.

See https://github.com/Matroska-Org/matroska-specification/pull/39#issuecomment-236765370

dericed:
Since EBML Elements default to minVer=1, then a version 0 EBML would require Elements that would be specifically defined for verison 0.

I like the way that FFV1 handles versioning with experimental files. Where each version (from 3 onward) is experimental until the subversion is marked as stable. So for instance in development of version 3 of ffv1, all subversions where experimental until subversion 4. Presumably all subversions of version 4 of FFV1 are experimental since no subversion is stable yet.

To handle this, we could consider adding EBMLDocTypeSubVersion, defaulting to `1`.

robUx4:
or `EBMLDocTypeVersionExperimental`

dericed:
Hi, just noting here that we discussed this issue on a recent cellar call. https://datatracker.ietf.org/meeting/interim-2018-cellar-01/materials/minutes-interim-2018-cellar-01-201805292000. We discussed that there are two issues here:
1. Is `docTypeVersion=0` valid?
2. How to handle experimental versions of an EBML Schema.

I propose that `docTypeVersion=0` is invalid since the `docTypeVersion` element is itself minver=1. Thus docTypeVersion is undefined if docTypeVersion=0, so it's just illogical. I'll add a PR to say that docTypeVersion must have a range of `>0`.

Re handling experimental versions of an EBML Schema, I like the way FFV1 handles this by having a subVersion, then at some point in each version's development a particular subversion is noted as the first non-experimental subversion of that version. For instance ffv1 3.0, 3.1, 3.2, and 3.3 are experimental whereas 3.4 (version 3, subversion 4) is defined as stable.

robUx4:
I agree with both proposals.

robUx4:
Also `EBMLDocTypeVersionExperimental` may be too restrictive, you may have updates on a proposal for an unstable version and don't want to end up with just a bool telling it's OK or not. So the subversion sounds better.  IMO it should be not 0. So if it's present it means the files has some extra fields that may be unknown voluntarily (compared to the plain "0" version).

mcr:
I believe that you have concensus to mark =0 as invalid.

dericed:
[...]

PR #178, opened 2018-06-01
set docType and readDocType to not 0 range

dericed:
As these elements are defined as `minver=1`, a `0` value is invalid.

partly addresses https://github.com/Matroska-Org/ebml-specification/issues/157


retokromer (approved):

robUx4 (approved):
LGTM

robUx4:
OK based on the discussion in #157

PR #179, opened 2018-06-07
IANA considerations for Element ID allocation

mcr:
This is my proposed text on Element ID allocation.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -785,11 +785,11 @@ encoded using the VINT mechanism described in
 section `Variable Sized Integer` can be between one and five bytes
 long. Five byte long Element IDs are possible only if declared in the header.
 
-One byte Element IDs are numbers between 0x01 and 0x7e. These items are valuable
+One byte Element IDs are numbers between 0x01 and 0x7E. These items are valuable
 because they are short, and need to be used for commonly repeated elements.
 Values from 1 to 126 are to be allocated according to RFC Required.
 
-Two byte Element IDs are numbers between 0x7f and 0x3FFE.
+Two byte Element IDs are numbers between 0x7F and 0x3FFE.
 Numbers may be allocated within this range according to Specification Required.
 
 The numbers 0x3FFF and 0x4000 are RESERVED.
--- a/specification.markdown
+++ b/specification.markdown
@@ -800,7 +800,3 @@ Other Four Byte Element IDs may be allocated by First Come First Served (see [@!
 The numbers 0xFFFFFFF and 0x1000000 are RESERVED.
 
 Five Byte Element IDs (values from 0x10000001 upwards) are reserved for Experimental use: they may be used by anyone at any time, but there is no coordination.
-
-
-
-
--- a/specification.markdown
+++ b/specification.markdown
@@ -777,5 +777,44 @@ description: Used to void damaged data, to avoid unexpected behaviors when using
 
 # IANA Considerations
 
-This document has no IANA actions.
+This document creates a new IANA Registry called
+"CELLAR EBML Element ID Registry".
+
+Element IDs are described in section {{#element-id}}.  Element IDs are
+encoded using the VINT mechanism described in
+section {{#variable-sized-integer}} can be between one and five bytes
+long. Five byte long Element IDs are possible only if declared in the header.
+
+One byte Element IDs are numbers between 1 and 126. These items are valuable
+because they short, and need to be used for commonly repeated elements.
+Values from 1 to 126 are to be allocated according to RFC Required.
+
+Two byte Element IDs are numbers between 127 and 16382.
+Numbers may be allocated within this range according to Specification Required.
+
+Three byte Element IDs are numbers between 16383 and 2097150.
+Numbers may be allocated within this range according to First Come First Served.
+
+Four byte Element IDs are numbers between 2097151 and 268435456.
+Four byte Element IDs are somewhat special in that they are useful for
+resynchronizing to major structures in the event of data corruption or
+loss.  As such four byte Element IDs are split into two categories.
+Four byte Element IDs whose lower three bytes (as encoded) would make
+printable 7-bit ASCII values may be allocated only Specification Required.
+Sequential allocation of values is not required: specifications SHOULD
+include a specific request, and are encouraged to do early allocations.
+
+To be clear about the above category: Four Byte Element IDs always start
+with hex 0x10 to 0x1F,  and that byte may be chosen so that the entire number
+has some desirable property, such as a specific CRC.  The other three bytes,
+when ALL having values between 33 (ASCII !) and 126 (ASCII ~), fall into
+this catgory.
+
+Other Four Byte Element IDs may be allocated by First Come First Served.
+
+Five Byte Element IDs (values from 268435457 upwards) are reserved for
+Experimental use.
+
+
+
 
--- a/specification.markdown
+++ b/specification.markdown
@@ -96,10 +96,6 @@ Side channel attacks could exploit:
 - Copies of `Identically Recurring Element` within a `Parent Element` that contain invalid `CRC-32 Elements`.
 - Use of `Void Elements`.
 
-# IANA Considerations
-
-This document has no IANA actions.
-
 # Structure
 
 `EBML` uses a system of `Elements` to compose an `EBML Document`. `EBML Elements` incorporate three parts: an `Element ID`, an `Element Data Size`, and `Element Data`. The `Element Data`, which is described by the `Element ID`, includes either binary data, one or many other `EBML Elements`, or both.
@@ -778,3 +774,8 @@ minOccurs: 0
 type: Binary
 
 description: Used to void damaged data, to avoid unexpected behaviors when using damaged data. The content is discarded. Also used to reserve space in a sub-element for later use.
+
+# IANA Considerations
+
+This document has no IANA actions.
+
--- a/specification.markdown
+++ b/specification.markdown
@@ -777,50 +777,29 @@ description: Used to void damaged data, to avoid unexpected behaviors when using
 
 # IANA Considerations
 
-This document creates a new IANA Registry called
-"CELLAR EBML Element ID Registry".
+This document creates a new IANA Registry called "CELLAR EBML Element ID Registry".
 
-Element IDs are described in section `Element Id`.  Element IDs are
-encoded using the VINT mechanism described in
-section `Variable Sized Integer` can be between one and five bytes
-long. Five byte long Element IDs are possible only if declared in the header.
+Element IDs are described in section `Element Id`.  Element IDs are encoded using the VINT mechanism described in section `Variable Sized Integer` can be between one and five bytes long. Five byte long Element IDs are possible only if declared in the header.
 
-One byte Element IDs are numbers between 0x01 and 0x7E. These items are valuable
-because they are short, and need to be used for commonly repeated elements.
-Values from 1 to 126 are to be allocated according to RFC Required.
+One byte Element IDs are numbers between 0x01 and 0x7E. These items are valuable because they are short, and need to be used for commonly repeated elements. Values from 1 to 126 are to be allocated according to RFC Required.
 
-Two byte Element IDs are numbers between 0x7F and 0x3FFE.
-Numbers may be allocated within this range according to Specification Required.
+Two byte Element IDs are numbers between 0x7F and 0x3FFE. Numbers may be allocated within this range according to Specification Required.
 
 The numbers 0x3FFF and 0x4000 are RESERVED.
 
-Three byte Element IDs are numbers between 0x4001 and 0x1F_FFFE.
-Numbers may be allocated within this range according to First Come First Served
-(see [@!RFC8126])
+Three byte Element IDs are numbers between 0x4001 and 0x1F_FFFE. Numbers may be allocated within this range according to First Come First Served (see [@!RFC8126])
 
 The numbers 0x1F_FFFF and 0x20_0000 are RESERVED.
 
-Four byte Element IDs are numbers between 0x200_0001 and 0xFFF_FFFE.
-Four byte Element IDs are somewhat special in that they are useful for
-resynchronizing to major structures in the event of data corruption or
-loss.  As such four byte Element IDs are split into two categories.
-Four byte Element IDs whose lower three bytes (as encoded) would make
-printable 7-bit ASCII values may be allocated only Specification Required.
-Sequential allocation of values is not required: specifications SHOULD
-include a specific request, and are encouraged to do early allocations.
+Four byte Element IDs are numbers between 0x200_0001 and 0xFFF_FFFE. Four byte Element IDs are somewhat special in that they are useful for resynchronizing to major structures in the event of data corruption or loss.  As such four byte Element IDs are split into two categories.  Four byte Element IDs whose lower three bytes (as encoded) would make printable 7-bit ASCII values may be allocated only Specification Required.  Sequential allocation of values is not required: specifications SHOULD include a specific request, and are encouraged to do early allocations.
 
-To be clear about the above category: Four Byte Element IDs always start
-with hex 0x10 to 0x1F,  and that byte may be chosen so that the entire number
-has some desirable property, such as a specific CRC.  The other three bytes,
-when ALL having values between 0x21 (33, ASCII !) and 0x7e (126, ASCII ~), fall into
-this catgory.
+To be clear about the above category: Four Byte Element IDs always start with hex 0x10 to 0x1F,  and that byte may be chosen so that the entire number has some desirable property, such as a specific CRC.  The other three bytes, when ALL having values between 0x21 (33, ASCII !) and 0x7e (126, ASCII ~), fall into this catgory.
 
 Other Four Byte Element IDs may be allocated by First Come First Served (see [@!RFC8126]).
 
 The numbers 0xFFF_FFFF and 0x100_0000 are RESERVED.
 
-Five Byte Element IDs (values from 0x1000_0001 upwards) are reserved for
-Experimental use: they may be used by anyone at any time, but there is no coordination.
+Five Byte Element IDs (values from 0x1000_0001 upwards) are reserved for Experimental use: they may be used by anyone at any time, but there is no coordination.
 
 
 
--- a/specification.markdown
+++ b/specification.markdown
@@ -780,22 +780,27 @@ description: Used to void damaged data, to avoid unexpected behaviors when using
 This document creates a new IANA Registry called
 "CELLAR EBML Element ID Registry".
 
-Element IDs are described in section {{#element-id}}.  Element IDs are
+Element IDs are described in section `Element Id`.  Element IDs are
 encoded using the VINT mechanism described in
-section {{#variable-sized-integer}} can be between one and five bytes
+section `Variable Sized Integer` can be between one and five bytes
 long. Five byte long Element IDs are possible only if declared in the header.
 
-One byte Element IDs are numbers between 1 and 126. These items are valuable
-because they short, and need to be used for commonly repeated elements.
+One byte Element IDs are numbers between 0x01 and 0x7e. These items are valuable
+because they are short, and need to be used for commonly repeated elements.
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+
+This document creates a new IANA Registry called
+"CELLAR EBML Element ID Registry".
+
+Element IDs are described in section {{#element-id}}.  Element IDs are
+encoded using the VINT mechanism described in
+section {{#variable-sized-integer}} can be between one and five bytes
+long. Five byte long Element IDs are possible only if declared in the header.
+
+One byte Element IDs are numbers between 1 and 126. These items are valuable
```
dericed: Maybe better to say something like "The VINT_DATA of one-byte Element IDs are numbers between 1 and 126." since the number references only the VINT_DATA part. Otherwise a one byte Element ID would only be in the range of 0x81...0xFE.

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+
+# IANA Considerations
+
+This document creates a new IANA Registry called
+"CELLAR EBML Element ID Registry".
+
+Element IDs are described in section {{#element-id}}.  Element IDs are
+encoded using the VINT mechanism described in
+section {{#variable-sized-integer}} can be between one and five bytes
+long. Five byte long Element IDs are possible only if declared in the header.
```
dericed: Maybe reference EBMLMaxIDLength specifically here.

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
 type: Binary
 
 description: Used to void damaged data, to avoid unexpected behaviors when using damaged data. The content is discarded. Also used to reserve space in a sub-element for later use.
+
+# IANA Considerations
+
+This document creates a new IANA Registry called
+"CELLAR EBML Element ID Registry".
+
+Element IDs are described in section {{#element-id}}.  Element IDs are
```
dericed: for style, `Element ID` and derivatives (plurals) are bickticked here. Also we usually don't force line breaks but let the make process add them in.

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+encoded using the VINT mechanism described in
+section {{#variable-sized-integer}} can be between one and five bytes
+long. Five byte long Element IDs are possible only if declared in the header.
+
+One byte Element IDs are numbers between 1 and 126. These items are valuable
+because they short, and need to be used for commonly repeated elements.
+Values from 1 to 126 are to be allocated according to RFC Required.
+
+Two byte Element IDs are numbers between 127 and 16382.
```
dericed: suggest to use comma delimiters at 1000's increments in numbers

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+Numbers may be allocated within this range according to Specification Required.
+
+Three byte Element IDs are numbers between 16383 and 2097150.
+Numbers may be allocated within this range according to First Come First Served.
+
+Four byte Element IDs are numbers between 2097151 and 268435456.
```
dericed: 268,435,454 ... (right?).
Also perhaps these min/max values should be integrated into the table in the {{#element-id}} section

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+Four byte Element IDs are somewhat special in that they are useful for
+resynchronizing to major structures in the event of data corruption or
+loss.  As such four byte Element IDs are split into two categories.
+Four byte Element IDs whose lower three bytes (as encoded) would make
+printable 7-bit ASCII values may be allocated only Specification Required.
```
dericed: Should define the ascii range specifically here. For instance the lower three bytes of the EBML Element are 0x45DFA3 which is extended ascii for Eß£ (~'ebml')

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+this catgory.
+
+Other Four Byte Element IDs may be allocated by First Come First Served.
+
+Five Byte Element IDs (values from 268435457 upwards) are reserved for
+Experimental use.
+
+
+
+
```
dericed: excessive line break parade

specification.markdown
dericed: ok i see you define a range below

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+Four byte Element IDs are numbers between 2097151 and 268435456.
+Four byte Element IDs are somewhat special in that they are useful for
+resynchronizing to major structures in the event of data corruption or
+loss.  As such four byte Element IDs are split into two categories.
+Four byte Element IDs whose lower three bytes (as encoded) would make
+printable 7-bit ASCII values may be allocated only Specification Required.
+Sequential allocation of values is not required: specifications SHOULD
+include a specific request, and are encouraged to do early allocations.
+
+To be clear about the above category: Four Byte Element IDs always start
```
dericed: perhaps this paragraph and the prior could be merged

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+with hex 0x10 to 0x1F,  and that byte may be chosen so that the entire number
+has some desirable property, such as a specific CRC.  The other three bytes,
+when ALL having values between 33 (ASCII !) and 126 (ASCII ~), fall into
+this catgory.
+
+Other Four Byte Element IDs may be allocated by First Come First Served.
```
dericed: how?

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+
+Five Byte Element IDs (values from 268435457 upwards) are reserved for
+Experimental use.
```
dericed: that only leaves two available non-experimental five byte element ids.

mcr:
I am responding to comments in github by Dave Rice.

I have fixed the backtick for references and the , for thousands.
If you want to merge the paragraphs, I will do that once I'm finished
editing, because I find it annoying to edit that way (and it makes for
less understable diffs).

I have fixed the off-by-one on 268,435,456.

> Also perhaps these min/max values should be integrated into the table in
> the {{#element-id}} section

That's a good idea.

> Should define the ascii range specifically here. For instance the lower
> three bytes of the EBML Element are 0x45DFA3 which is extended ascii for
> Eß£ (~'ebml')

If we permit other than 7-bit US-ASCII to be reserved, then the discussion
will then be latin-1,latin-2,windows-XXXX, or UTF-8.  If you want 0x45DFA3,
then it's also available by First Come First Served, it's just not reserved.

> perhaps this paragraph and the prior could be merged

my opinion is that it more understandable this way.

> +Other Four Byte Element IDs may be allocated by First Come First Served.
> how?

Please see https://tools.ietf.org/html/rfc8126#section-4.4

>+Five Byte Element IDs (values from 268435457 upwards) are reserved for
>+Experimental use.
>
>that only leaves two available non-experimental five byte element ids.

I don't think it leaves any: it's all experimental.
Do we really need that any Five Byte Element IDs in specifications?







-- 
Michael Richardson <mcr+IETF@sandelman.ca>, Sandelman Software Works
 -= IPv6 IoT consulting =-

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+Four byte Element IDs are somewhat special in that they are useful for
+resynchronizing to major structures in the event of data corruption or
+loss.  As such four byte Element IDs are split into two categories.
+Four byte Element IDs whose lower three bytes (as encoded) would make
+printable 7-bit ASCII values may be allocated only Specification Required.
+Sequential allocation of values is not required: specifications SHOULD
+include a specific request, and are encouraged to do early allocations.
+
+To be clear about the above category: Four Byte Element IDs always start
+with hex 0x10 to 0x1F,  and that byte may be chosen so that the entire number
```
dericed: double space in this line

specification.markdown
dericed: Is 'may' correct here? As opposed to an RFC2119 SHOULD.

mcr:
> On Jun 7, 2018, at 1:08 PM, Michael Richardson <mcr+ietf@sandelman.ca> wrote:
> 
> I am responding to comments in github by Dave Rice.
> 
> I have fixed the backtick for references and the , for thousands.
> If you want to merge the paragraphs, I will do that once I'm finished
> editing, because I find it annoying to edit that way (and it makes for
> less understable diffs).
> 
> I have fixed the off-by-one on 268,435,456.
> 
>> Also perhaps these min/max values should be integrated into the table in
>> the {{#element-id}} section
> 
> That's a good idea.
> 
>> Should define the ascii range specifically here. For instance the lower
>> three bytes of the EBML Element are 0x45DFA3 which is extended ascii for
>> Eß£ (~'ebml')
> 
> If we permit other than 7-bit US-ASCII to be reserved, then the discussion
> will then be latin-1,latin-2,windows-XXXX, or UTF-8.  If you want 0x45DFA3,
> then it's also available by First Come First Served, it's just not reserved.

Keeping it to the 7-bit ascii range is fine with me. Nudging Steve for comment as he’s the one that originated use of element id easter eggs outside of the 0x20 to 0x7E range.

>> perhaps this paragraph and the prior could be merged
> 
> my opinion is that it more understandable this way.
> 
>> +Other Four Byte Element IDs may be allocated by First Come First Served.
>> how?
> 
> Please see https://tools.ietf.org/html/rfc8126#section-4.4

I suggest including an "[@!RFC8126]” at the end of this sentence to create a normative reference to RFC8126.

>> +Five Byte Element IDs (values from 268435457 upwards) are reserved for
>> +Experimental use.
>> 
>> that only leaves two available non-experimental five byte element ids.
> 
> I don't think it leaves any: it's all experimental.

This may be related to the off by one issue. If "Five Byte Element IDs (values from 268435457 upwards) are reserved for Experimental use.”
Both 268435456 and 268435457 require a five element id, so in my first reading it seems like this sentence was saying that:

268435456, reservation status unknown
268435457, reservation status unknown
>268435457, reserved

> Do we really need that any Five Byte Element IDs in specifications?

I don’t need five byte element IDs in this specification, but a user defining an EBML Document Type should be able to read this document and understand if and how they can create their own identifiers at any length.

Dave

mcr:
> On Jun 13, 2018, at 11:23 AM, Michael Richardson <mcr@sandelman.ca> wrote:
> 
> 
> Dave Rice <dave@dericed.com> wrote:
>>> I don't think it leaves any: it's all experimental.
> 
>> This may be related to the off by one issue. If "Five Byte Element IDs
>> (values from 268435457 upwards) are reserved for Experimental use.”
>> Both 268435456 and 268435457 require a five element id, so in my first
>> reading it seems like this sentence was saying that:
> 
>> 268435456, reservation status unknown
>> 268435457, reservation status unknown
> 
> I would mark them as simply RESERVED, that way nobody can get the encoding
> wrong.
> 
>>> Do we really need that any Five Byte Element IDs in specifications?
> 
>> I don’t need five byte element IDs in this specification, but a user
>> defining an EBML Document Type should be able to read this document and
>> understand if and how they can create their own identifiers at any
>> length.
> 
> The First Come First Served space is usually easiest.
> 
> If you feel that the text needs to be friendly, or explain things better to
> non-IETF types, that's completely reasonable.
> My goal is to write text that *IANA* will understand, and that represents the
> rough consensus of the group.
> So, again, are there any objections to the essential plan?

The essential plan sounds fine with me. Looking forward to reviewing an updated pull request.
Dave

ablwr:
On Wed, Jun 13, 2018 at 11:26 AM, Michael Richardson <
notifications@github.com> wrote:

>
> Dave Rice <dave@dericed.com> wrote:
> >> I don't think it leaves any: it's all experimental.
>
> > This may be related to the off by one issue. If "Five Byte Element IDs
> > (values from 268435457 upwards) are reserved for Experimental use.”
> > Both 268435456 and 268435457 require a five element id, so in my first
> > reading it seems like this sentence was saying that:
>
> > 268435456, reservation status unknown
> > 268435457, reservation status unknown
>
> I would mark them as simply RESERVED, that way nobody can get the encoding
> wrong.
>
> >> Do we really need that any Five Byte Element IDs in specifications?
>
> > I don’t need five byte element IDs in this specification, but a user
> > defining an EBML Document Type should be able to read this document and
> > understand if and how they can create their own identifiers at any
> > length.
>
> The First Come First Served space is usually easiest.
>
> If you feel that the text needs to be friendly, or explain things better to
> non-IETF types, that's completely reasonable.
> My goal is to write text that *IANA* will understand, and that represents
> the
> rough consensus of the group.
> So, again, are there any objections to the essential plan?
>
>
>
No objection here.

Ashley

retokromer:
I am fine with this.

specification.markdown
```diff
@@ -778,3 +774,47 @@ minOccurs: 0
...
+This document creates a new IANA Registry called
+"CELLAR EBML Element ID Registry".
+
+Element IDs are described in section {{#element-id}}.  Element IDs are
+encoded using the VINT mechanism described in
+section {{#variable-sized-integer}} can be between one and five bytes
+long. Five byte long Element IDs are possible only if declared in the header.
+
+One byte Element IDs are numbers between 1 and 126. These items are valuable
+because they short, and need to be used for commonly repeated elements.
```
robUx4: they *are* short ? (missing verb)

robUx4:
> So, again, are there any objections to the essential plan?

No, thank you for doing this !

robUx4:
> I suggest including an "[@!RFC8126]” at the end of this sentence to create a normative reference to RFC8126.

I second that.

specification.markdown
```diff
@@ -780,22 +780,27 @@ description: Used to void damaged data, to avoid unexpected behaviors when using
 This document creates a new IANA Registry called
 "CELLAR EBML Element ID Registry".
 
-Element IDs are described in section {{#element-id}}.  Element IDs are
+Element IDs are described in section `Element Id`.  Element IDs are
```
retokromer: nit: double space

specification.markdown
```diff
@@ -780,22 +780,27 @@ description: Used to void damaged data, to avoid unexpected behaviors when using
...
 long. Five byte long Element IDs are possible only if declared in the header.
 
-One byte Element IDs are numbers between 1 and 126. These items are valuable
-because they short, and need to be used for commonly repeated elements.
+One byte Element IDs are numbers between 0x01 and 0x7e. These items are valuable
+because they are short, and need to be used for commonly repeated elements.
 Values from 1 to 126 are to be allocated according to RFC Required.
 
-Two byte Element IDs are numbers between 127 and 16382.
+Two byte Element IDs are numbers between 0x7f and 0x3FFE.
```
retokromer: 0x7F

specification.markdown
```diff
@@ -807,13 +812,15 @@ include a specific request, and are encouraged to do early allocations.
 To be clear about the above category: Four Byte Element IDs always start
 with hex 0x10 to 0x1F,  and that byte may be chosen so that the entire number
 has some desirable property, such as a specific CRC.  The other three bytes,
-when ALL having values between 33 (ASCII !) and 126 (ASCII ~), fall into
+when ALL having values between 0x21 (33, ASCII !) and 0x7e (126, ASCII ~), fall into
```
retokromer: 0x7E

specification.markdown
```diff
@@ -780,22 +780,27 @@ description: Used to void damaged data, to avoid unexpected behaviors when using
...
-Element IDs are described in section {{#element-id}}.  Element IDs are
+Element IDs are described in section `Element Id`.  Element IDs are
 encoded using the VINT mechanism described in
-section {{#variable-sized-integer}} can be between one and five bytes
+section `Variable Sized Integer` can be between one and five bytes
 long. Five byte long Element IDs are possible only if declared in the header.
 
-One byte Element IDs are numbers between 1 and 126. These items are valuable
-because they short, and need to be used for commonly repeated elements.
+One byte Element IDs are numbers between 0x01 and 0x7e. These items are valuable
```
retokromer: 0x7E

specification.markdown
```diff
@@ -778,3 +774,33 @@ minOccurs: 0
...
+
+The numbers 0x1F_FFFF and 0x20_0000 are RESERVED.
+
+Four byte Element IDs are numbers between 0x200_0001 and 0xFFF_FFFE. Four byte Element IDs are somewhat special in that they are useful for resynchronizing to major structures in the event of data corruption or loss.  As such four byte Element IDs are split into two categories.  Four byte Element IDs whose lower three bytes (as encoded) would make printable 7-bit ASCII values may be allocated only Specification Required.  Sequential allocation of values is not required: specifications SHOULD include a specific request, and are encouraged to do early allocations.
+
+To be clear about the above category: Four Byte Element IDs always start with hex 0x10 to 0x1F,  and that byte may be chosen so that the entire number has some desirable property, such as a specific CRC.  The other three bytes, when ALL having values between 0x21 (33, ASCII !) and 0x7e (126, ASCII ~), fall into this catgory.
+
+Other Four Byte Element IDs may be allocated by First Come First Served (see [@!RFC8126]).
+
+The numbers 0xFFF_FFFF and 0x100_0000 are RESERVED.
```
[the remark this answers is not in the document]
mcr: Please see: https://tools.ietf.org/html/rfc8126. Invalid is not a valid instruction to IANA.

specification.markdown
```diff
@@ -778,3 +774,33 @@ minOccurs: 0
...
+
+Element IDs are described in section `Element ID`.  Element IDs are encoded using the VINT mechanism described in section (#variable-size-integer) can be between one and five bytes long. Five byte long Element IDs are possible only if declared in the header.
+
+The VINT Data value of one-byte Element IDs MUST be between 0x01 and 0x7E. These items are valuable because they are short, and need to be used for commonly repeated elements. Values from 1 to 126 are to be allocated according to RFC Required.
+
+The VINT Data value of two-byte Element IDs MUST be between 0x007F and 0x3FFE. Numbers MAY be allocated within this range according to Specification Required.
+
+The numbers 0x3FFF and 0x4000 are RESERVED.
+
+The VINT Data value of three-byte Element IDs MUST be between 0x4001 and 0x1F_FFFE. Numbers may be allocated within this range according to First Come First Served (see [@!RFC8126])
```
robUx4: What's with the underscore on values bigger than 16 bits ?

robUx4:
Shouldn't the IANA section also list the `path` of the element within the `DocType` and also what `DocTypeVersion` it applies to ? An ID could theoretically be used in different places. And we need to know when an ID became valid.

robUx4:
Actually there might be confusion here (at least on my part). I think there will be an EBML IANA registry and a Matroska IANA registry. EBML deals with elements added to the EBML header (no `DocType` involved, but the `EBMLVersion`). Actually that should also cover the list of known `DocType` within this registry. And EBML ID's should rather go in an EBML spec. The Matroska registry would deal with the ID's within that `DocType`.

mcr:
Steve Lhomme <notifications@github.com> wrote:
    > Shouldn't the IANA section also list the path of the element within the
    > DocType and also what DocTypeVersion it applies to ? An ID could
    > theoretically be used in different places. And we need to know when an ID
    > became valid.

This is only about allocation of Element IDs, and I understood that this
space was flat (not dependant on any other context). I understand that not
every ElementID is valid in every DocType, but that there is no plan to
overload ElementIDs between DocTypes or DocTypeVersions.

robUx4:
So how do we address the issue between unrelated DocType that will likely use the same IDs ? IMO Matroska should have its own registry. It's up to other formats to handle the way they add new elements. On the other hand what is really unique in EBML is not the IDs for each `DocType` (they are not) but the actual `DocType` strings. For example the most common are "matroska" and "webm". These should be publicly known and not reused for anything else.

So it seems this paragraph about how to define IDs is necessary, to be reused by `DocType` specs but shouldn't actually register anything at the EBML level. We need one paragraph for `DocType` strings.

dericed:
Seems like I can commit to @mcr's pull request so I removed the underscores in the hex expressions and the extra line breaks.

dericed:
I'm also completely fine with the proposed changes. Thanks!

dericed:
Going to squash and merge before the meeting today if no objections.

mbunkus:
Please do, yes.

PR #180, opened 2018-06-24
add DocTypeSubVersion to handle experimental elements

robUx4:
This is a proposal to handle experimental elements without having to define a `DocType` value each time.

Just as I wrote this I realized a more flexible approach could be the following:
DocTypeExtension
  DocTypeExtensionName
  DocTypeExtensionVersion

This allows different workgroups working on different versions of an element subset. It may even allow listing these elements in there in the future. The DocTypeExtension would be multiple.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -747,6 +747,24 @@ type: Unsigned Integer
 
 description: The minimum `DocType` version an `EBML Reader` has to support to read this `EBML Document`. The value of the `DocTypeReadVersion Element` MUST be less than or equal to the value of the `DocTypeVersion Element`.
 
+### DocTypeSubVersion Element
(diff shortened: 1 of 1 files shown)


retokromer:
>Just as I wrote this I realized a more flexible approach could be the following:
>DocTypeExtension
>DocTypeExtensionName
>DocTypeExtensionVersion

+1

dericed:
If used, consider adding 'Default: 0'

mbunkus:
I strongly dislike this approach and favor the one proposed in #181 where I also explained my concerns about #180.

robUx4:
OK, everyone seems to agree this is bad. Closing.

PR #185, opened 2018-08-28
rename bytes to octets to match the rest of the specs

robUx4:
Since that's what we used everywhere. It's also clearer than byte, which may (mostly in the past) represent more than 8 bits.


mbunkus:
I approve. However, I just noticed that Byte was capitalized in a couple of places (e.g. "Four Byte Element IDs") but not in others. Maybe lower-case those while you're at it.

dericed:
Thx, I still approve. I suggest to squash and merge.

PR #186, opened 2018-08-28
bump to version 06

dericed:
just leaving this here to follow the end of today's merges


robUx4:
IMO you can do the bump once #177 is resolved.
#181 is not really meant for the current EBML, it would be a last minute addition.

dericed:
Maybe let's add in https://github.com/Matroska-Org/ebml-specification/pull/189 then merge and handle the other open PR's later.

Issue #183, opened 2018-07-02
set iana considerations for docType values


dericed:
see https://github.com/Matroska-Org/ebml-specification/pull/179#issuecomment-400191417

PR #188, opened 2018-08-28
add doctype registry to iana considerations

dericed:
to resolve https://github.com/Matroska-Org/ebml-specification/issues/183

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -789,6 +789,8 @@ DocType values are described in [this section](#doctype).  DocTypes are ASCII st
 
 The use of ASCII corresponds to the types and code already in use, the value is not meant to be visible to the user.
 
+DocType string values of "matroska" and "webm" are reserved for pre-existing formats.
+
 ## CELLAR EBML Element ID Registry
 
 This document creates a new IANA Registry called "CELLAR EBML Element ID Registry".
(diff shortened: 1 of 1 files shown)


robUx4:
IMO one doesn't go with the other. It was never the goal to restrict the IDs from one EBML format in other EBML formats. Otherwise we'll run out of Class A IDs very quickly.

IMO for EBML we only need a registry of `DocType`, for example `"matroska"` and `"webm"` are the 2 current entries.

Then Matroska has its own IANA registry for its EBML IDs.

dericed:
@robUx4, that comment seems more relevant to https://github.com/Matroska-Org/ebml-specification/pull/179 which was already merged. We can discuss in the call with Michael.

robUx4:
Yes I came too late before the merge happened. But IMO it can be discussed here since you're adding a significant part that was missing.

robUx4:
I updated the PR with the First Come First Served rule, moving it first and reserving "matroska"/"webm".

dericed:
LGTM, I can't approve officially in github since I started the PR, but I approve here in the comments.

mbunkus:
As we all approve and agree so much, I'll merge it.

PR #191, opened 2018-08-30
Make the ID registry about the EBML Header only

robUx4:
And add initial values.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -813,7 +813,7 @@ This document creates a new IANA Registry called "CELLAR EBML Element ID Registr
 
 Element IDs are described in section `Element ID`. Element IDs are encoded using the VINT mechanism described in section (#variable-size-integer) can be between one and five octets long. Five octet long Element IDs are possible only if declared in the header.
 
-This IANA Registry only applies to `Elements` contained in the `EBML Header` and global `Elements` that are either valid in the `EBML Header` or the whole `EBML Document`. `Elements` only found in the `EBML Body` have their own set of independent `Element IDs` and are not part of this IANA Registry.
+This IANA Registry only applies to `Elements` contained in the `EBML Header`. `Elements` only found in the `EBML Body` have their own set of independent `Element IDs` and are not part of this IANA Registry.
 
 The VINT Data value of one-octet Element IDs MUST be between 0x01 and 0x7E. These items are valuable because they are short, and need to be used for commonly repeated elements. Values from 1 to 126 are to be allocated according to RFC Required.
 
@@ -837,15 +837,15 @@ Five octet Element IDs (values from 0x10000001 upwards) are reserved for Experim
 
 ID Values found in this document are assigned as initial values as follows:
 
- ID        | Name
-----------:|:-------------------------
-0x1A45DFA3 | EBML Head
-0x4286     | EBML Version
-0x42F7     | EBML Read Version
-0x42F2     | EBML Max ID Length
-0x42F3     | EBML Max Size Length
-0x4282     | EBML DocType
-0x4287     | EBML DocType Version
-0x4285     | EBML DocType Read Version
+ ID        | Element Name
+----------:|:-------------------------------
+0x1A45DFA3 | EBML
+0x4286     | EBMLVersion
+0x42F7     | EBMLReadVersion
+0x42F2     | EBMLMaxIDLength
+0x42F3     | EBMLMaxSizeLength
+0x4282     | DocType
+0x4287     | DocTypeVersion
+0x4285     | DocTypeReadVersion
 0xBF       | CRC-32
 0xEC       | Void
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -811,7 +811,9 @@ DocType string values of "matroska" and "webm" are reserved for pre-existing for
 
 This document creates a new IANA Registry called "CELLAR EBML Element ID Registry".
 
-Element IDs are described in section `Element ID`.  Element IDs are encoded using the VINT mechanism described in section (#variable-size-integer) can be between one and five octets long. Five octet long Element IDs are possible only if declared in the header.
+Element IDs are described in section `Element ID`. Element IDs are encoded using the VINT mechanism described in section (#variable-size-integer) can be between one and five octets long. Five octet long Element IDs are possible only if declared in the header.
+
+This IANA Registry only applies to `Elements` contained in the `EBML Header`, not the `Elements` used in the `EBML Body` which have their own set of independent `Element IDs`.
```
dericed: Should this registry apply to Global Elements as well (Void and CRC-32)?

specification.markdown
```diff
@@ -833,3 +835,17 @@ The numbers 0xFFFFFFF and 0x1000000 are RESERVED.
...
+ ID        | Name
+----------:|:-------------------------
+0x1A45DFA3 | EBML Head
+0x4286     | EBML Version
+0x42F7     | EBML Read Version
+0x42F2     | EBML Max ID Length
+0x42F3     | EBML Max Size Length
+0x4282     | EBML DocType
+0x4287     | EBML DocType Version
+0x4285     | EBML DocType Read Version
```
dericed: I suggest to use the same name as elsewhere in the document (no spaces and `EBML` as the name for 0x1A45DFA3).

robUx4: For each of them ? IMO the registry is meant to be more user readable. Whereas those names are the ones found in the Path to avoid any ambiguity.

specification.markdown
robUx4: Indeed, we need to explain what happens to global elements.

specification.markdown
dericed: The camelcasing maintains readability. At any rate, not a strong objection.

robUx4:
Ping @mbunkus

specification.markdown
```diff
@@ -811,7 +811,9 @@ DocType string values of "matroska" and "webm" are reserved for pre-existing for
 
 This document creates a new IANA Registry called "CELLAR EBML Element ID Registry".
 
-Element IDs are described in section `Element ID`.  Element IDs are encoded using the VINT mechanism described in section (#variable-size-integer) can be between one and five octets long. Five octet long Element IDs are possible only if declared in the header.
+Element IDs are described in section `Element ID`. Element IDs are encoded using the VINT mechanism described in section (#variable-size-integer) can be between one and five octets long. Five octet long Element IDs are possible only if declared in the header.
+
+This IANA Registry only applies to `Elements` contained in the `EBML Header` and global `Elements` that are either valid in the `EBML Header` or the whole `EBML Document`. `Elements` only found in the `EBML Body` have their own set of independent `Element IDs` and are not part of this IANA Registry.
```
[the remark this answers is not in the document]
robUx4: It's actually anywhere that doesn't exclude the EBML Header. Or in other words, that include the EBML Header.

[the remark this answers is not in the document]
robUx4: So I'll just remove the end of the sentence.

mbunkus:
Hmm, you removed any mention of "global elements" now. Not sure if that's the right thing to do; CRC-32 and Void are global, after all. It might be clearer to mention globals again :)

mcr:
Moritz Bunkus <notifications@github.com> wrote:
    > Hmmm we do have some discrepancies regarding element naming. You're
    > either talking about the EBML Document Type in general or the DocType
    > Element, but not the EBML DocType Element you include in this list.
    > Similar, you're talking about CRC-32 Elements elsewhere in the document
    > but in this list you use CRC-32 (which would match the rest of the
    > document but not the rest of the table; otherwise you'd have to use
    > EBML CRC-32.

Yes, we want to make it consistent.
Usually there is a third column in a registry which says which document.

+ ID        | Name                     | Document
+----------:|:-------------------------+---------
+0x1A45DFA3 | EBML Head                | [THIS DOCUMENT] section X
+0x4286     | EBML Version             | [THIS DOCUMENT] section X
+0x42F7     | EBML Read Version        | [THIS DOCUMENT] section X
+0x42F2     | EBML Max ID Length       | [THIS DOCUMENT] section X
+0x42F3     | EBML Max Size Length     | [THIS DOCUMENT] section X
+0x4282     | EBML DocType             | [THIS DOCUMENT] section X
+0x4287     | EBML DocType Version     | [THIS DOCUMENT] section X
+0x4285     | EBML DocType Read Version| [THIS DOCUMENT] section X

    > I'd prefer to use the element names as they're used in the headings
    > that describe them: EBML, EBMLVersion and EBMLMaxIDLength, but
    > DocTypeVersion, CRC-32 and Void. That would be precise and easy enough
    > to cross-reference.

No need to guess, put the reference right in.

robUx4:
Added a reference to the document and a wording that includes global elements explicitly, the ones found in the EBML Header.

PR #192, opened 2018-09-05
bump to 07

dericed:
lot of recent activity, so here's a version bump


dericed:
suggest resolving https://github.com/Matroska-Org/ebml-specification/pull/193 first

robUx4:
We can merge that so that local files don't produce a -06 version on top of the frozen -06 one.
In general once a version is published we should bump the version right away.

mcr:
Steve Lhomme <notifications@github.com> wrote:
    > In general once a version is published we
    > should bump the version right away.

yes, do that to make rfcdiff easier.

PR #193, opened 2018-09-05
escape xml elements in sectional headers

dericed:
otherwise the value is dropped altogether when the output docs are rendered


robUx4:
LGTM

robUx4:
LGTM

PR #200, opened 2019-01-19
add versioning paragraphs

robUx4:


specification.markdown
```diff
@@ -331,6 +331,18 @@ All data of an `EBML Document` following the `EBML Header` is the `EBML Body`. T
 
 An `EBML Stream` is a file that consists of one or many `EBML Documents` that are concatenated together. An occurrence of a `EBML Header` at the `Root Level` marks the beginning of an `EBML Document`.
 
+# EBML Versioning
+
+An `EBML Document` handles 2 different versions: the version of the `EBML Header` and the version of the `EBML Body`. Both versions are meant to be backward compatible.
```
dericed: There seems a minor mismatch here. I think I'd prefer to not use the term `EBML Body` here, but prefer `EBML Document` since the value has long been called `EBML Document Version` rather than `EBML Body Version`.

robUx4: In fact it *is* the version of the `EBML Body`. The definition of the `EBML Document` is:
> A datastream comprised of only two components, an `EBML Header` and an `EBML Body`.

The `EBML Header` has it version and the `EBML Body` has its own. They can be changed separately, which couldn't be the case if the `EBMLDocTypeVersion` was covering both parts.

dericed: I agree, but then perhaps the first sentence of the `EBML Document Version` should be "The version of the EBML Body is found in EBMLDocTypeVersion". Worried though that contextualizing `EBMLDocTypeVersion` as the version of the Body rather than the Document will require more changes.

robUx4: Agreed, I'll add it to the patch although it's technically a different issue.

Issue #197, opened 2019-01-07
clarity security consideration


dericed:
From https://mailarchive.ietf.org/arch/msg/cellar/ZeTFBMZ_acBakwYFGZlhM-XJ5yU

Regarding https://tools.ietf.org/html/draft-ietf-cellar-ebml-08#section-3

> §3: Can you offer guidance on the specific impacts of the mentioned attacks, and how to mitigate them? (If there is no way to mitigate an attack, it's okay to say that.)

PR #202, opened 2019-01-19
Error mitigation

robUx4:
Separate OK errors and not OK errors.

I'm not sure what to do with the side channel errors.


robUx4:
Fixes partially #197

robUx4:
> 
> 
> As for side channel attacks: I'd simply say that a reader MAY use the elements anyway at its discretion, just like with the other possible attacks (e.g. by ignoring trailing 0 bytes in string elements meant for displaying).

I added a line for that.

PR #203, opened 2019-01-19
explain why it's like UTF-8 and why it's not and why UTF-8

robUx4:


robUx4:
OK, I am fine with removing it altogether.

PR #210, opened 2019-01-23
bump to version 09

dericed:


dericed:
Been 4 months since a version bump

retokromer:
Can https://github.com/Matroska-Org/ebml-specification/pull/200 not be resolved before?

dericed:
@retokromer I stalled 18 hours but don't consider that #200 is blocking to the version bump

retokromer:
It’s not a blocking issue. It would just having been «nice» to have resolved it in the new version. No more, no less.

dericed:
Sorry @retokromer, but work is moving quickly in response to the AD review and @mcr recommended more versions rather than waiting for moments of empty PR lists.

PR #212, opened 2019-01-27
An EBML Document should only have EBML Elements

robUx4:
But it MAY has some extra data as seen in the streaming case.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -288,7 +288,7 @@ If the Element to be changed is a Descendant Element of any Master Element that 
 
 # EBML Document
 
-An EBML Document is comprised of only two components, an EBML Header and an EBML Body. An EBML Document MUST start with an EBML Header that declares significant characteristics of the entire EBML Body. An EBML Document consists of EBML Elements and MUST NOT contain any data that is not part of an EBML Element.
+An EBML Document is comprised of only two components, an EBML Header and an EBML Body. An EBML Document MUST start with an EBML Header that declares significant characteristics of the entire EBML Body. An EBML Document consists of EBML Elements and SHOULD NOT contain any data that is not part of an EBML Element.
 
 ## EBML Header
 


dericed:
Maybe we could say: An EBML Document consists one EBML Header Element immediately followed by one Root Element and does not contain any other data.

robUx4:
I think the issue is rather how normative that is and what happens to data not fitting the EBML structure.

dericed:
I disagree with this change. With MUST, then the demuxer can expect that the Root Element immediately follows the EBML Header, but the change to SHOULD implies that `<EBML/>{junk}<Segment/>` is not breaking a mandate.

robUx4:
This is from the review of Ben Campbell of draft 08

```
§11: "An "EBML Document" consists of "EBML Elements" and MUST
NOT contain any data that is not part of an "EBML Element"."

That contradicts text in §8.7 that says you can do this in streaming use cases.
```

And indeed I know GStreamer used to (maybe not anymore) start sending the data even in the middle of an element when live streaming.

dericed:
Maybe better wording is needed but I don't see the contradiction.

For me, EBML Document is not an EBML Element. An EBML Document is series of an EBML Header Element (an EBML Element) immediately followed by a Root Element (another EBML Element), so an EBML Document is basically 2 EBML Elements of particular meaning (header and root). What I want to avoid is to allow to consider a data stream such as `<EBML Header/>LOTS OF GARBARGE<Root Element>`. To me `<EBML Header/><Root Element>` complies with section 11, even if that `Root Element` contains garbage data as it may be allowed to do so.

tl;dr I'm ok with garbage in an `EBML Element` but would prefer not to allow non-EBML Element data with an `EBML Document`. If non-EBML Element data was allowed in an EBML Document, then `junk<EBML Element/>junk<Root Element/>junk` would be allowed and we'd never be able to say what is the beginning of end of an EBML Document within a stream and we'd lose the presumption that an EBML Header should be followed by the Root Element.

robUx4:
OK but that's not how GStreamer used to do it for live streaming. It would send the EBML Header, then the Segment "headers", the Cluster "header" and then parts of blocks it's already sending to other clients. That's the garbage we are talking about.

I think nowadays we shouldn't allow this. It's fine for a Matroska-aware server to hold data until a new keyframe block to send to the client. That's what VLC does when Chromecasting for example. Hence proposal #254 instead of this one.

robUx4:
Closing as I think we should use #254

PR #214, opened 2019-01-27
Path example

robUx4:


specification.markdown
```diff
@@ -403,6 +403,8 @@ The `VariableParentOccurrence` part is interpreted as an ABNF Variable Repetitio
 
 If the path contains an `EBMLPathAtomRecursive` part, the `EBML Element` can occur within itself recursively (see the [recursive attribute](#recursive)).
 
+As an example, the `path` for `EBMLMaxIDLength`, according to its [definition](#ebmlmaxidlength-element), translates to "1*1(\EBML\EBMLMaxIDLength)". Meaning it's found insided the `EBML` at least once and at most once.
+
```
retokromer: Meaning it's found inside the `EBML` exactly once.

robUx4: Yes, it's explaining the 1 and the 1... Which is not a good example...

retokromer (approved):

PR #216, opened 2019-01-27
mention the range for printable ASCII

robUx4:


specification.markdown
```diff
@@ -876,7 +876,7 @@ The VINT Data value of three-octet Element IDs MUST be between 0x4001 and 0x1FFF
...
 
-Four octet Element IDs are numbers between 0x2000001 and 0xFFFFFFE. Four octet Element IDs are somewhat special in that they are useful for resynchronizing to major structures in the event of data corruption or loss. As such four octet Element IDs are split into two categories. Four octet Element IDs whose lower three octets (as encoded) would make printable 7-bit ASCII values MUST be allocated by the "Specification Required" policy. Sequential allocation of values is not required: specifications SHOULD include a specific request, and are encouraged to do early allocations.
+Four octet Element IDs are numbers between 0x2000001 and 0xFFFFFFE. Four octet Element IDs are somewhat special in that they are useful for resynchronizing to major structures in the event of data corruption or loss. As such four octet Element IDs are split into two categories. Four octet Element IDs whose lower three octets (as encoded) would make printable 7-bit ASCII values (0x20 to 0x7f) MUST be allocated by the "Specification Required" policy. Sequential allocation of values is not required: specifications SHOULD include a specific request, and are encouraged to do early allocations.
```
retokromer: We should decide if we use A-F or a-f for hex, and use it systematically in one same way in the whole document.

robUx4: I fixed the remaining printable `0x7F`

mcr:
0x7F is RUBOUT (aka Delete), and it is not a printable character.
Generally one also wants to avoid 0x20 (32-space).

robUx4:
> 0x7F is RUBOUT (aka Delete), and it is not a printable character. Generally one also wants to avoid 0x20 (32-space).

Should they be given another name than printable ?

PR #219, opened 2019-02-11
Relax EBML Header length restrictions

mkver:
libavformat (ffmpeg's multiplexer) has always coded the size field of the EBML element on eight bytes for all Matroska and Webm files it was used to mux. The current requirement of coding every size in the header on four bytes or less would therefore invalidate lots of files which makes it unacceptable.

As this requirement was stated twice, I have removed one occurence of it. The description of EBML Body doesn't repeat the restrictions imposed by EBMLMaxSize/IDLength either.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -82,7 +82,7 @@ The `Element ID` and `Element Data Size` are both encoded as a `Variable Size In
 
 Each `Variable Size Integer` begins with a `VINT_WIDTH` which consists of zero or many zero-value bits. The count of consecutive zero-values of the `VINT_WIDTH` plus one equals the length in octets of the `Variable Size Integer`. For example, a `Variable Size Integer` that starts with a `VINT_WIDTH` which contains zero consecutive zero-value bits is one octet in length and a `Variable Size Integer` that starts with one consecutive zero-value bit is two octets in length. The `VINT_WIDTH` MUST only contain zero-value bits or be empty.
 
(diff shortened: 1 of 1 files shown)


robUx4:
This change seems more complex and less readable. Also in libavformat only the `0x1A45DFA3` has a size of 8 or all elements in the EBML header ? I think it would be better to allow only the EBML Header Master Element to have a size of 8 if possible. The more data, the more damage is possible to the header.

robUx4:
I just checked a file recently created with lavf and only the `EBML Header Master Element` (`0x1A45DFA3`) has a length size of 8. The elements inside the EBML Header have "optimized" sizes.

So IMO only the `EBML Header Master Element` should be allowed to have a bigger length, not the elements inside.

mkver:
Yes, only master elements are possibly written with too many octets than needed by current libavformat. But I am not so sure about older versions. I will take a look at the git history and if possible update this PR. (It's ok to force-push, isn't it?)

robUx4:
You can either push a new commit on top or push force, it's up to you.

If the backtick changes go through before your PRs you may have to update your commits anyway. And a rebase would be nicer.

mkver:
libavformat has always written the length of the other header elements on the lowest number of octets possible. I also checked the oldest mkvmerge version I could find (0.4.2) and it also writes the elements with as few bytes as possible. Therefore your propsal is feasible. I have already updated the PR.

robUx4:
LGTM

PR #226, opened 2019-02-17
unbacktick internal names

### LINKED ISSUE #225: Remove some double quotes of defined names
At least the ones in upper case. Or maybe it's the backtick to RFC text translation that does that ?

As reported by Robert Sparks
https://mailarchive.ietf.org/arch/msg/cellar/gCFVwpnS79yZFsGk6yXucB5sMn0

robUx4:
It doesn't produce readable RFC text (or we would need to customize xml2rfc to
keep a nice emphasis in the HTML version).

Fixes #225

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -360,11 +360,11 @@ The version attribute is REQUIRED within the `<EBMLSchema>` Element.
 
 Each `<element>` defines one EBML Element through the use of several attributes that are defined in [EBML Schema Element Attributes](#ebmlschema-attributes). EBML Schemas MAY contain additional attributes to extend the semantics but MUST NOT conflict with the definitions of the `<element>` attributes defined within this document.
 
-The `<element> nodes contain a description of the meaning and use of the EBML Element stored within one or more `<documentation>` sub-elements and zero or one `<restriction>` sub-element. All `<element>` nodes MUST be sub-elements of the `<EBMLSchema>`.
+The `<element>` nodes contain a description of the meaning and use of the EBML Element stored within one or more `<documentation>` sub-elements and zero or one `<restriction>` sub-element. All `<element>` nodes MUST be sub-elements of the `<EBMLSchema>`.
 
 ### \<element> Attributes
 
-Within an EBML Schema the `<element> uses the following attributes to define an EBML Element`:
+Within an EBML Schema the `<element>` uses the following attributes to define an EBML Element:
 
 #### name
 
@@ -506,7 +506,7 @@ The maxver attribute is OPTIONAL. If the maxver attribute is not present then th
 
 ### \<documentation> Element
 
-The `<documentation> element provides additional information about the EBML Element`.
+The `<documentation>` element provides additional information about the EBML Element.
 
 ### \<documentation> Attributes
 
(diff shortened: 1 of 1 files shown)


robUx4:
The backtick is kept for XML elements in the code. The double quoting in the TEXT output is actually welcome.

specification.markdown
```diff
@@ -1,100 +1,100 @@
 # Introduction
 
-`EBML`, short for Extensible Binary Meta Language, specifies a binary and octet (byte) aligned format inspired by the principle of XML (a framework for structuring data).
+EBML, short for Extensible Binary Meta Language, specifies a binary and octet (byte) aligned format inspired by the principle of XML (a framework for structuring data).
```
retokromer: Should the first occurrence not be «quoted», as definition?

robUx4: Fixed in the next commits

robUx4: In the `Notation and Conventions` section, not the main text.

specification.markdown
```diff
@@ -160,7 +160,7 @@ Octet Length | Possible Value Range
 7            | 0 to 2^49-2
 8            | 0 to 2^56-2
 
-If the length of `Element Data` equals `2^(n*7)-1` then the octet length of the `Element Data Size` MUST be at least `n+1`. This rule prevents an `Element Data Size` from being expressed as a reserved value. The following table clarifies this rule by showing a valid and invalid expression of an `Element Data Size` with a `VINT_DATA` of 127 (which is equal to 2^(1*7)-1) and 16,383 (which is equal to 2^(2*7)-1).
+If the length of Element Data equals `2^(n*7)-1 then the octet length of the Element Data Size MUST be at least n+1`. This rule prevents an Element Data Size from being expressed as a reserved value. The following table clarifies this rule by showing a valid and invalid expression of an Element Data Size with a VINT_DATA of 127 (which is equal to 2^(1*7)-1) and 16,383 (which is equal to 2^(2*7)-1).
```
mkver: The backticked block here includes two forumulae and the statement in between.

robUx4: OK I will merge your patch and then update this one.

dericed:
@robUx4 I'll try to resolve this one for re-review.

specification.markdown
dericed: There's still two backticks left in here which join together two formula and text into a part that shouldn't be singularly encapsulated.

specification.markdown
```diff
@@ -271,112 +271,112 @@ After  | 0x3B4040   | 0x82              | 0x6869       | 0xEC80
...
 
-An `EBML Schema` MUST declare exactly one `EBML Element` at `Root Level` (referred to as the `Root Element`) that occurs exactly once within an `EBML Document`. The `Void Element` MAY also occur at `Root Level` but is not a `Root Element` (see [the definition of the `Void Element`](#void-element)).
+An EBML Schema MUST declare exactly one EBML Element at Root Level (referred to as the Root Element) that occurs exactly once within an EBML Document. The Void Element MAY also occur at Root Level but is not a Root Element (see [the definition of the Void Element](#void-element)).
 
-The `EBML Schema` MUST document all Elements of the `EBML Body`. The `EBML Schema` does not document `Global Elements` that are defined by this document (namely the `Void Element` and the `CRC-32 Element`).
+The EBML Schema MUST document all Elements of the EBML Body. The EBML Schema does not document Global Elements that are defined by this document (namely the Void Element and the CRC-32 Element).
 
 The EBML Schema` MUST NOT use the Element ID `0x1A45DFA3` which is reserved for the EBML Header for resynchronization purpose.
```
dericed: oops, stray backtick in this line

robUx4: Fixed in the next version

specification.markdown
```diff
@@ -399,162 +399,162 @@ PathMinOccurrence        = 1*DIGIT
...
-A `lang` attribute which is set to the [@!RFC5646] value of the language of the element's documentation.
+A lang attribute which is set to the [@!RFC5646] value of the language of the element's documentation.
 
-The `lang` attribute is OPTIONAL.
+The lang attribute is OPTIONAL.
 
 #### purpose
 
-A `purpose` attribute distinguishes the meaning of the documentation. Values for the `<documentation>` sub-element's `purpose` attribute MUST include one of the following: `definition`, `rationale`, `usage notes`, and `references`.
+A purpose attribute distinguishes the meaning of the documentation. Values for the <documentation> sub-element's purpose attribute MUST include one of the following: 'definition`, 'rationale`, 'usage notes`, and 'references`.
```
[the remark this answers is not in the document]
robUx4: My mistake, fixed more generally with elements that are found in XML

robUx4:
Can we agree on merging this ? Because it's a PITA to maintain and keep writing new patches in a form that will need to be edited.

specification.markdown
robUx4: Fixed

dericed:
agreed

dericed:
fyi, after merging this, I bumped the version 10 merged to follow it (rather the precede)

PR #227, opened 2019-02-21
Fix table showing the number of IDs in each EBML Class

mkver:
The table is currently wrong: The valid IDs in class B correspond to the VINT_DATA in the range of 2^7 - 1 to 2^14 - 2 (inclusive) so that there are 2^14 - 2 - (2^7 - 1) + 1 = 2^14 - 2^7 of them; it's similar for the other classes apart from A.

Notice that the current numbers can't be explained with 0x3FFF and 0x4000 etc. being reserved, because then the formulae were still wrong (it would need -2 for class C and D, too, instead of just for class A) and because they are only reserved for the header (and if they were reserved everywhere, then this would need to be mentioned in the section about EBML IDs, because otherwise the table would be unintelligible).

I could have written the amount in the form 2^14 - 2 - (2^7 - 1) + 1, but this would probably wrap, so I didn't do it.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -134,10 +134,10 @@ The octet length of an `Element ID` determines its `EBML Class`.
 
 EBML Class | Octet Length | Number of Possible Element IDs
 :---------:|:------------:|:------------------------------
-Class A    | 1            | 2^7  - 2        =         126
-Class B    | 2            | 2^14 - 2^7  - 1 =      16,255
-Class C    | 3            | 2^21 - 2^14 - 1 =   2,080,767
-Class D    | 4            | 2^28 - 2^21 - 1 = 266,338,303
+Class A    | 1            | 2^7  - 2    =         126
+Class B    | 2            | 2^14 - 2^7  =      16,256
+Class C    | 3            | 2^21 - 2^14 =   2,080,768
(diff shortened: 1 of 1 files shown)


robUx4:
IMO it would be easier to state the values in the valid range, the math is then a subtraction:
```
Class A: 0x81 to 0xFE = 126
Class B: 0x4001 to 0x7FFE = 16,381
Class C: 0x200001 to 0x3FFFFE = 2,097,149
Class D: 0x10000001 to 0x1FFFFFFE = 268,435,453
```

mkver:
Your table is wrong: 0x4001, 0x200001 and 0x10000001 are not coded on the least number of octets possible, hence they are invalid IDs. The real table is

Your table is wrong: 0x4001, 0x200001 and 0x10000001 are not coded on the least number of octets possible, hence they are invalid IDs (and classes B-D are off by one even if we ignore that). The real table is:
```
Class A: 0x81 to 0xFE = 126
Class B: 0x407F to 0x7FFE = 16,256
Class C: 0x203FFF to 0x3FFFFE = 2,080,768
Class D: 0x101FFFFF to 0x1FFFFFFE = 266,338,304
```

dericed:
superseded by https://github.com/Matroska-Org/ebml-specification/pull/230

PR #228, opened 2019-02-21
Escape "*"

mkver:
This is done in order to distinguish "*" (the multiplication sign) from being interpreted as emphasis by markdown. Addresses one of the issues raised in Robert Sparks' review.

This commit also unbackticks two other formulae (one very similar to this one) to bring them into line with other formulae.

I don't know if formulae should be backticked or not; the relevant part of #226 is wrong.


dericed:
will merge today unless objection

robUx4:
LGTM

PR #241, opened 2019-02-27
Show the number of usable bits in a VINT

robUx4:
We don't need to tell how many values are possible

All 1's may or may not be part of what is considered usable but that's a
different issue.


dericed (approved):

retokromer (approved):

dericed (dismissed):
I revoke my approval. After the patch, the preceding paragraph references a size column that is no longer in the table.

robUx4:
Updated the bogus paragraph.

retokromer (approved):

dericed (approved):

PR #242, opened 2019-02-27
Benign security conditions cleaning

robUx4:
Also mention it depends on how strict the parser wants to be.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -862,8 +862,10 @@ An EBML Document that has the following issues may still be handled by the EBML 
 - Invalid Element IDs that are longer than the limit stated in the EBMLMaxIDLength Element of the EBML Header.
 - Invalid Element IDs that are not encoded in the shortest-possible way.
 - Invalid Element Data Size values that are longer than the limit stated in the EBMLMaxSizeLength Element of the EBML Header.
-- Invalid Element IDs comprised of reserved values.
-- Usage of 0x00 octets in EBML Elements with a string type.
+
+Invalid Element IDs comprised of reserved values MAY be accepted as valid EBML IDs in order to skip such elements.
+
+EBML Elements with a string type may contain extra data after the first 0x00. These data SHOULD be discarded according to the (#terminating-elements) rules.
 
 An EBML Reader may discard some or all data if the following errors are found in the EBML Document:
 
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -843,13 +843,15 @@ EBML itself does not offer any kind of security and does not provide confidentia
...
-An EBML Document that has the following issues may still be handled by the EBML Reader and the data accepted as such:
+An EBML Document that has the following issues may still be handled by the EBML Reader and the data accepted as such depending on how strict it wants to enforce the EBML Header rules:
 
 - Invalid Element IDs that are longer than the limit stated in the EBMLMaxIDLength Element of the EBML Header.
 - Invalid Element IDs that are not encoded in the shortest-possible way.
-- Invalid Element IDs comprised of reserved values.
 - Invalid Element Data Size values that are longer than the limit stated in the EBMLMaxSizeLength Element of the EBML Header.
-- Usage of 0x00 octets in EBML Elements with a string type.
+
+Invalid Element IDs comprised of reserved values MAY be accepted as valid EBML IDs in order to skip such elements.
```
robUx4: Note that if such reserved elements are ever used, they are not reserved anymore and parsed as such in new parsers. Old parsers will skip these data properly.

dericed: do you need the word invalid here? Are any of the reserved values also invalid Element IDs?

specification.markdown
```diff
@@ -843,13 +843,15 @@ EBML itself does not offer any kind of security and does not provide confidentia
...
+
+EBML Elements with a string type may contain extra data after the first 0x00. These data SHOULD be discarded according to the (#terminating-elements) rules.
```
dericed: why not MUST?

specification.markdown
robUx4: how about `legit` or `allowed` ? They are not invalid so they might actually be used for something in the future.

specification.markdown
robUx4: Technically one could hide content in there. But I guess that's their problem... So `MUST` it is.

specification.markdown
dericed: Perhaps `Element IDs that are unknown to the EBML Reader...`

specification.markdown
dericed: no strong opinion on MUST or SHOULD. Will approve as is.

PR #243, opened 2019-02-27
rewrite the range boundary format section

robUx4:
- allowing the combination of lower+upper boundaries
- forbidding the x-y range format if y is negative
- adding examples for each

It cover all the cases possible for numbers.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -451,9 +451,9 @@ For an exclusive lower boundary the `>` sign is used and the `>=` sign is used f
 
 For an exclusive upper boundary the `<` sign is used and the `<=` sign is used for an inclusive upper boundary. For example `<-2` meaning the Element value MUST be less than -2 or `<=10` meaning the Element value MUST be less than or equal to the 10.
 
-The lower and upper bounds can be combined into an expression to form a closed boundary. The lower boundary coming first followed by the upper boundary. For example `>3 <= 20` means the Element value MUST be greater than 3 and less than or equal to 20.
+The lower and upper bounds can be combined into an expression to form a closed boundary. The lower boundary coming first followed by the upper boundary, separated by a comma. For example `>3,<= 20` means the Element value MUST be greater than 3 and less than or equal to 20.
 
-A special form of lower and upper bounds using the `-` separator is possible, meaning the Element value MUST be greater than or to the first value and MUST be less than or equal to the second value. For example `1-10` is equivalent to `>=1 <=10`. If the upper boundary is negative, only the latter form MUST be used.
+A special form of lower and upper bounds using the `-` separator is possible, meaning the Element value MUST be greater than or to the first value and MUST be less than or equal to the second value. For example `1-10` is equivalent to `>=1,<=10`. If the upper boundary is negative, only the latter form MUST be used.
 
 
 #### length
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -443,16 +443,18 @@ The range attribute is OPTIONAL. If the range attribute is not present then any
...
 
-- x-y where x and y are integers or floats and y MUST be greater than x, meaning that the value MUST be greater than or equal to x and less than or equal to y. x MUST be less than y.
-- `>x where x` is an integer or float, meaning that the value MUST be greater than x.
-- `>=x where x` is an integer or float, meaning that the value MUST be greater than or equal to x.
-- `<x where x` is an integer or float, meaning that the value MUST be less than x.
-- `<=x where x` is an integer or float, meaning that the value MUST be less than or equal to x.
-- x where x is an integer or float, meaning that the value MUST be equal x.
+To set a fixed value for the range, the value is used as the attribute value. For example `1234` means the EBML element always has the value 1234. The value can be prefixed with `not` to indicate that the set value MUST not be used for that Element. For example `not 1234` means the Element can use all values of its type except 1234.
```
dericed: Via rfc2119, `MUST NOT` should have the 'not' capitalized as well.

dericed: should `always has the value 1234` be `MUST have a value of 1234`. The latter infers that the element is mandated to have that value whereas the former simply infers that it does have that value.

dericed: I suggest that "with \`not\`" be "with \`not \`" (to include the space)

specification.markdown
```diff
@@ -443,16 +443,18 @@ The range attribute is OPTIONAL. If the range attribute is not present then any
...
+
+For an exclusive lower boundary the `>` sign is used and the `>=` sign is used for an inclusive lower boundary. For example `>3` meaning the Element value MUST be greater than 3 or `>= 0x1p+0` meaning the Element value MUST be greater than or equal to the floating value 1.0, see [textual expression of floats](#textual-expression-of-floats).
+
+For an exclusive upper boundary the `<` sign is used and the `<=` sign is used for an inclusive upper boundary. For example `<-2` meaning the Element value MUST be less than -2 or `<=10` meaning the Element value MUST be less than or equal to the 10.
+
+The lower and upper bounds can be combined into an expression to form a closed boundary. The lower boundary coming first followed by the upper boundary. For example `>3 <= 20` means the Element value MUST be greater than 3 and less than or equal to 20.
```
dericed: Here there is a space delimiting between `<=` and `20` whereas in the next example there is no space between `<=` and `10`. I suggest to remove that space, but also perhaps we need to state a specific delimiter. Since <space> is meaningless in other examples such as `not 1234` perhaps a comma is a better delimiter, such as `>3,<=10`.

dericed: Perhaps the use of the `not ` should be clarified more. For instance, from this a reader could consider that `not <3` is a valid range.

specification.markdown
robUx4: > should `always has the value 1234` be `MUST have a value of 1234`. The latter infers that the element is mandated to have that value whereas the former simply infers that it does have that value.

This is an example so I don't think it should have anything normative.

robUx4: > I suggest that "with `not`" be "with `not `" (to include the space)

the backtick already produces a double quote, so it would end up like 
> The value can be prefixed "with ""not" to indicate that...

specification.markdown
robUx4: > Perhaps the use of the `not ` should be clarified more. For instance, from this a reader could consider that `not <3` is a valid range.

Actually `not` only applies to the "fixed value" it cannot be used for the other cases.

PR #252, opened 2019-03-10
misc security consideration (short) explanations

robUx4:


specification.markdown
```diff
@@ -869,18 +869,18 @@ An EBML Reader may discard some or all data if the following errors are found in
...
 
-- The semantic equivalence of the same string stored in a String Element or UTF-8 Element with and without zero-bit padding.
-- The semantic equivalence of VINT_DATA within Element Data Size with two different lengths due to left-padding zero bits.
-- Data contained within a Master Element which is not itself part of an EBML Element.
-- Extraneous copies of Identically Recurring Element.
-- Copies of Identically Recurring Element within a Parent Element that contain invalid CRC-32 Elements.
-- Use of Void Elements.
+- The semantic equivalence of the same string stored in a String Element or UTF-8 Element with and without zero-bit padding, making comparison at the semantic level invalid.
+- The semantic equivalence of VINT_DATA within Element Data Size with two different lengths due to left-padding zero bits, making comparison at the semantic level invalid.
+- Data contained within a Master Element which is not itself part of an EBML Element can trigger incorrect parsing behavior in EBML Readers.
```
dericed: All data within a Master Element is part of an EBML Element (the same referenced Master Element). I suggest `which is not itself part of a Child Element`.

specification.markdown
```diff
@@ -869,18 +869,18 @@ An EBML Reader may discard some or all data if the following errors are found in
...
+- Extraneous copies of Identically Recurring Element, making parsing unnecessarily slow to the point of not being usable.
+- Copies of Identically Recurring Element within a Parent Element that contain invalid CRC-32 Elements. EBML Readers not checking the CRC-32 might use the bogus version of the element.
+- Use of Void Elements which could be used to hide content or create bogys resynchronzation points seen by some EBML Reader and not others.
```
dericed: `bogus`?

dericed: Would this comment impact the use of attachments as well? For instance if a Matroska was stored as an attachment within a Matroska file, the nest Matroska could cause false resynchronization.

specification.markdown
```diff
@@ -869,18 +869,18 @@ An EBML Reader may discard some or all data if the following errors are found in
 
 - Invalid Element Data Size values (e.g. extending the length of the EBML Element beyond the scope of the Parent Element; possibly triggering access-out-of-bounds issues).
 - Very high lengths in order to force out-of-memory situations resulting in a denial of service, access-out-of-bounds issues etc.
-- Missing EBML Elements that are mandatory and have no declared default value.
+- Missing EBML Elements that are mandatory in a Master Element and have no declared default value, making the semantic invalid at that Master Element level.
 - Usage of invalid UTF-8 encoding in EBML Elements of UTF-8 type (e.g. in order to trigger access-out-of-bounds or buffer overflow issues).
-- Usage of invalid data in EBML Elements with a date type.
+- Usage of invalid data in EBML Elements with a date type, trigerring bogus date accesses.
```
dericed: Are you referring to use of NaN or infinite values in date floating points? We could use more specific in the date definition and say that the date MUST be a finite number and that non-finite numbers as dates are a security issue.

specification.markdown

specification.markdown
robUx4: Nested Matroska data could indeed cause problems of resynchronization. But that's not the same issue as Void hiding data on purpose. Otherwise any binary data you put in Matroska could be misinterpreted at some point. 
I think Void is different because it can be found anywhere but not be handled properly everywhere if a parser hardcodes so possible location for Void.

specification.markdown
robUx4: Nope, dates are signed integers. But they can refer to dates way back which wouldn't make sense to seek to (for example).

specification.markdown
robUx4: Replaced "bogus" with mismatching CRC-32

PR #254, opened 2019-03-31
Don't allow invalid EBML in streaming anymore

robUx4:
The bogus data recovery mechanism of the player still apply if old streams are
found.


robUx4:
Replaces #212 so we always have a strict format (and trigger bogus data checks in all cases the same way).

robUx4:
Merged with typo fixed

Issue #239, opened 2019-02-26
Define the CRC-32 Element earlier in the document


robUx4:
Reported-By: Robert Sparks <rjsparks@nostrum.com>

robUx4:
I'd rather move the paragraphs `Terminating Elements` and `Guidelines for Updating Elements` after the elements are actually defined. Probably after `Considerations for Reading EBML Data`.

dericed:
resolved by https://github.com/cellar-wg/ebml-specification/pull/256?

robUx4:
Not per se, but the flow is better IMO.
I suggest we close it and reopen if anyone complains.

Issue #237, opened 2019-02-26
Start the document with an introduction


robUx4:
Add a diagram of the main structure of an EBML document (with a tree ?)

Reported-By: Robert Sparks <rjsparks@nostrum.com>

PR #256, opened 2019-05-08
Move some general paragraphs after the elements definition

robUx4:
"Terminating Elements" and "Guidelines for Updating Elements" should come after
"Considerations for Reading EBML Data".

There should still be a more general/less technical introduction at the beginning (#237)

Solves #239


JeromeMartinez:
> Solves #239

I don't get it, CRC is still defined late, isn't it?

dericed:
Updated to resolve conflicts. I kept the sections that @robUx4 moved before Security Considerations but just after the new Forward/Backward Compatibility section.

Issue #238, opened 2019-02-26
Add an example showing how Unknown Size elements work


robUx4:
With an example on how to find the end

Reported-By: Robert Sparks <rjsparks@nostrum.com>

robUx4:
Closing as it's fixed by #257

PR #257, opened 2019-05-08
Clarify the unknown size use

robUx4:
The text has been split, added some levels to make things easier to read.

I rewrote the text to detail the cases where an unknown element ends. Making sure it only ends with valid data and listing the various cases, ordered by likeliness of happening.

Since things are clearer, I don't think we need more examples as suggested by #238


specification.markdown
```diff
@@ -155,11 +155,21 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+
+- Any valid EBML Element according to the EBML Schema that is a Parent Element of the Unknown-Sized Element.
+- Any valid EBML Element according to the EBML Schema that is not a Descendant Element of the Unknown-Sized Element. For example a Top-Level Element.
```
mbunkus: I see two potential issues:

1. The first case is a special variant of the second case. Readers might wonder why we mention parent elements explicitly when the second rule covers them already.
2. Readers might wonder how global elements play into this. Maybe mention that global elements do not end an Unknown-Sized Element.

dericed: @mbunkus counter-proposal?

robUx4: 1. The second is too broad. It's only possible to have a non-descendant element if they share a parent at the same level. Or if it's a root (then no parent).
```
A>B>C>D
A>B>E     E after D is OK
A>B>F>G   G after D is not OK
A>B'       B' after D is OK
A'           A' after D is OK
[A']>B'     B' after D is not OK
```

I'll try to come up with better text and an example.

robUx4: 2. OK with the global elements not ending a unknown-length element

specification.markdown
```diff
@@ -172,6 +172,20 @@ The use of Unknown-Sized Elements allows for an EBML Element to be written and r
 - Any valid EBML Element according to the EBML Schema, Global Elements excluded, that is a root element.
 - The end of the EBML Document, either when reaching the end of the file or because a new EBML Header started.
 
+Consider an Unknown-Size Element which EBML path is `\root\level1\level2\elt`. Here are some possible and impossible way to end this element:
+
+EBML Path of next element          | Status
+-----------------------------------|------------------------------
+`\root\level1\level2`              | Ends the unknown-size element, as it a new Parent element
+`\root\level1`                     | Ends the unknown-size element, as it a new Parent element
+`\root`                            | Ends the unknown-size element, as it a new Parent element
```
mbunkus: These three lines miss their verbs: "…as it _is_ a new…"

robUx4:
Updated with the verb and more example of ending an unknown size element with a root.

specification.markdown
```diff
@@ -155,11 +155,39 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+
+EBML Path of next element          | Status
+-----------------------------------|------------------------------
+`\root\level1\level2`              | Ends the unknown-size element, as it is a new Parent element
+`\root\level1`                     | Ends the unknown-size element, as it is a new Parent element
+`\root`                            | Ends the unknown-size element, as it is a new root element
+`\root2`                           | Ends the unknown-size element, as it is a new root element
+`\root\level1\level2\other`        | Ends the unknown-size element, as they share the same parent
+`\root\level1\level2\elt\inside`   | Doesn't end the unknown-size element, it's a child of `elt`
+`\root\level1\level2\other\part`   | Isn't valid right after `elt`. `other` should be found first
```
dericed: From the description, it's unclear if you imply that `part` signifies the end of `elt`. I think not, but it's unclear. For instance if you have `part` as the first part of the data of `elt`, then is `elt` ended as considered empty, then `part` is considered misplaced?

specification.markdown
```diff
@@ -155,11 +155,39 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+`\root\level1\level2\elt\<global>` | Global Element is valid, it's a child of `elt`
+`\root\level1\level2\<global>`     | Global Element is invalid, it can only be a child of `elt` while parsing `elt`
```
dericed: I'm confused by this wording. The global element here seems to be placed in a valid way, but I think you're trying to see that a global can't be used to consider where the end of `elt` is. Actually in these two cases, it seems impossible to distinguish the parent of the global element.

specification.markdown
```diff
@@ -155,11 +155,39 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+- The end of the EBML Document, either when reaching the end of the file or because a new EBML Header started.
+
+Consider an Unknown-Size Element which EBML path is `\root\level1\level2\elt`. Here are some possible and impossible way to end this element:
+
+EBML Path of next element          | Status
+-----------------------------------|------------------------------
+`\root\level1\level2`              | Ends the unknown-size element, as it is a new Parent element
+`\root\level1`                     | Ends the unknown-size element, as it is a new Parent element
+`\root`                            | Ends the unknown-size element, as it is a new root element
+`\root2`                           | Ends the unknown-size element, as it is a new root element
```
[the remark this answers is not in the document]
dericed: So is 'root2' supposed to be a new Root Element, since as in two concatenated EBML Documents? Perhaps an xpath style of `root[2]` would be more intuitive than root2.

specification.markdown
dericed: Sorry, @robUx4, still lost on how to interpret this. I think if you are reading `elt` (unknown size) and a Global Element is found, it is unknown if the Global Element is a Child Element or a new Element that follows the end of the elt element. Is there a way to clarify?

dericed: so maybe we say "Global Element is valid, but it is indecipherable if the Global Element is a Child Element of `elt` or an Element that follows the end of the `elt` Element."

specification.markdown
```diff
@@ -155,11 +155,39 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+
+The use of Unknown-Sized Elements allows for an EBML Element to be written and read before the size of the EBML Element is known. Unknown-Sized Element MUST NOT be used or defined unnecessarily; however if the Element Data Size is not known before the Element Data is written, such as in some cases of data streaming, then Unknown-Sized Elements MAY be used. The end of an Unknown-Sized Element is determined by whichever comes first:
+
+- Any valid EBML Element according to the EBML Schema, Global Elements excluded, that is a Parent Element of the Unknown-Sized Element.
+- Any valid EBML Element according to the EBML Schema, Global Elements excluded, that is not a Descendant Element of the Unknown-Sized Element but share a common direct parent. For example a Top-Level Element.
+- Any valid EBML Element according to the EBML Schema, Global Elements excluded, that is a root element.
+- The end of the EBML Document, either when reaching the end of the file or because a new EBML Header started.
```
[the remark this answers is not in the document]
robUx4: > An EBML Element that is defined with an unknownsizeallowed attribute set to 1 MUST also have the unknownsizeallowed attribute of its Parent Element set to 1.

The parent element must have the flag to 1 as well. That doesn't mean it has to write an unknown size. But would it make sense to know the parent size but not the individual elements ? Maybe if rewriting the size afterwards in the parent but not in individual children. Or having the parent with a fixed size and then writing whatever inside without caring too much about the size. So I guess it's a possible use case.

specification.markdown
```diff
@@ -155,11 +155,39 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+
+Consider an Unknown-Size Element which EBML path is `\root\level1\level2\elt`. Here are some possible and impossible way to end this element:
+
+EBML Path of next element          | Status
+-----------------------------------|------------------------------
+`\root\level1\level2`              | Ends the unknown-size element, as it is a new Parent element
+`\root\level1`                     | Ends the unknown-size element, as it is a new Parent element
```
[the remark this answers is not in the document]
robUx4: Ancestor is not a defined term and `level2` is a "parent" of `\root\level1\level2\elt`, a new one. An Uncle element ? IMO that's more confusing than the text we have now.

specification.markdown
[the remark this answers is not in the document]
robUx4: All elements in the table are the `EBML Path of next element`. `\root` has the same ID as the root in `\root\level1\level2\elt` but it's a new element.

`\root2` is a different ID than the root in `\root\level1\level2\elt`.

specification.markdown
```diff
@@ -155,11 +155,39 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+`\root`                            | Ends the unknown-size element, as it is a new root element
+`\root2`                           | Ends the unknown-size element, as it is a new root element
+`\root\level1\level2\other`        | Ends the unknown-size element, as they share the same parent
```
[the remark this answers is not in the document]
robUx4: Yes we must not allow that. It would be impossible to know if `elt` is a child of itself or the same element at the same level.

specification.markdown
robUx4: It's more a reminder that you can't skip levels between elements. You have to have the proper parent first. 

It may be a bit confusing because the decision to end an Unknown-Sized Element happens when the ID of an new element is read, not when the whole next element is read.

In this case you can't get the ID for `part` if you don't have the ID for `other` first. This case is covered 2 lines above.

specification.markdown
robUx4: Yes, this case is not "invalid" as in "it cannot happen here", it just can't be interpreted this way because if you find a `Global Element` it's interpreted as the example above and no other way.

specification.markdown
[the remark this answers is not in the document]
robUx4: So this is covered by
> The end of the Parent Element with a known size has been reached.

specification.markdown
robUx4: I'll remove this line. It doesn't help understand anything not covered above.

specification.markdown
```diff
@@ -155,11 +155,40 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+`\root\level1`                     | Ends the Unknown-Sized Element, as it is a new Parent Element
+`\root`                            | Ends the Unknown-Sized Element, as it is a new Root Element
+`\root2`                           | Ends the Unknown-Sized Element, as it is a new Root Element
+`\root\level1\level2\other`        | Ends the Unknown-Sized Element, as they share the same parent
+`\root\level1\level2\elt`          | Ends the Unknown-Sized Element, as they share the same parent
+`\root\level1\level2\elt\inside`   | Doesn't end the Unknown-Sized Element, it's a child of `elt`
+`\root\level1\level2\elt\<global>` | Global Element is valid, it's a child of `elt`
+`\root\level1\level2\<global>`     | Global Element cannot be assumed to have this path, it can only be a child of `elt` while parsing `elt`
+
+
```
dericed: extra line break

specification.markdown
```diff
@@ -155,11 +155,40 @@ Class D    | 4      | 0x101FFFFF - 0x1FFFFFFE |   268,338,304
...
+:----------------------------------|:-----------------------------
+`\root\level1\level2`              | Ends the Unknown-Sized Element, as it is a new Parent Element
+`\root\level1`                     | Ends the Unknown-Sized Element, as it is a new Parent Element
+`\root`                            | Ends the Unknown-Sized Element, as it is a new Root Element
+`\root2`                           | Ends the Unknown-Sized Element, as it is a new Root Element
+`\root\level1\level2\other`        | Ends the Unknown-Sized Element, as they share the same parent
+`\root\level1\level2\elt`          | Ends the Unknown-Sized Element, as they share the same parent
+`\root\level1\level2\elt\inside`   | Doesn't end the Unknown-Sized Element, it's a child of `elt`
+`\root\level1\level2\elt\<global>` | Global Element is valid, it's a child of `elt`
+`\root\level1\level2\<global>`     | Global Element cannot be assumed to have this path, it can only be a child of `elt` while parsing `elt`
```
dericed: could add

```
\<global>
\root\<global>
\root\level1\<global>
```

Also now we should contextualize the <global> because now we define their path to say `where` they are global. So for instance if the global is defined to not be a child of `elt` but is permitted as a  child of one of elt's ancestors than it could be used to end `elt`.

specification.markdown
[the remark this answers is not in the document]
robUx4: Actually it should be called a Grand Parent. But do we really want to introduce new terms just for this ?

specification.markdown
robUx4: This should go in the "global element" section, not the "unknown size" section. The fact that an element cannot be found at certain level is unrelated to how you can end an unknown length element. Otherwise we're going to need a table for each case where an element is legal or not and how to interpret such ID, etc. This section should only focus on cases where you know an unknown length size element has ended.

robUx4: For example I could add:
> Global Element is valid, it's a child of `elt` (assuming <global> is valid at this path)

But then we would need to do the same for every line above. It seems obvious that it's ends the element because, on top of its position, it's actually a valid element. I'll try add some text before the table to explain that.

robUx4:
I transformed the table introduction into this. I hope it clarifies things:

> Consider an Unknown-Sized Element which EBML path is `\root\level1\level2\elt`. When reading a new Element ID, assuming the EBML Path of that new Element is valid, here are some possible and impossible ways that this new Element is ending `elt`:

robUx4:
@dericed I'm happy if you're happy

PR #262, opened 2019-05-26
make the purpose element mandatory in the documentation Schema element

robUx4:
Otherwise we need to define how to interpret the <documentation> tag if it's
not present.


dericed:
@robUx4 I added a second commit to make the corresponding change in the EBMLSchema.xsd document and then a third to validate the purpose vocabulary

PR #270, opened 2019-07-07
Add ebml extension element

dericed:
As discussed in the cellar meeting. This subelement would allow a way to store webm data, cppname, etc without mixed such extensions with the rest of the element definition.

So in Matroska's EBML Schema:

```xml
<element name="AttachmentLink" path="0*1(\Segment\Tracks\TrackEntry\AttachmentLink)" cppname="TrackAttachmentLink" id="0x7446" type="uinteger" maxOccurs="1" minver="1" maxver="3" webm="0" range="not 0">
  <documentation lang="en">The UID of an attachment that is used by this codec.</documentation>
</element>
```
would become
```xml
<element name="AttachmentLink" path="0*1(\Segment\Tracks\TrackEntry\AttachmentLink)" id="0x7446" type="uinteger" maxOccurs="1" minver="1" maxver="3" range="not 0">
  <documentation lang="en">The UID of an attachment that is used by this codec.</documentation>
  <extension cppname="TrackAttachmentLink" webm="0"/>
</element>
```

This makes it a lot easier to have a valid EBML Schema for Matroska.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -312,7 +312,7 @@ The version attribute is REQUIRED within the `<EBMLSchema>` Element.
 
 Each `<element>` defines one EBML Element through the use of several attributes that are defined in [EBML Schema Element Attributes](#ebmlschema-attributes). EBML Schemas MAY contain additional attributes to extend the semantics but MUST NOT conflict with the definitions of the `<element>` attributes defined within this document.
 
-The `<element>` nodes contain a description of the meaning and use of the EBML Element stored within one or more `<documentation>` sub-elements, optional `<implementation_note>` sub-elements, and zero or one `<restriction>` sub-element. All `<element>` nodes MUST be sub-elements of the `<EBMLSchema>`.
+The `<element>` nodes contain a description of the meaning and use of the EBML Element stored within one or more `<documentation>` sub-elements, followed by optional `<implementation_note>` sub-elements, followed by zero or one `<restriction>` sub-element, followed by optional `<extension>` sub-elements. All `<element>` nodes MUST be sub-elements of the `<EBMLSchema>`.
 
 ### \<element> Attributes
 
@@ -509,6 +509,18 @@ The label attribute is OPTIONAL.
 The value represents data that MAY be stored within the EBML Element.
 
 The value attribute is REQUIRED.
+
+### \<extension> Element
+
+ The `<extension>` element provides an unconstained element to contain information about the associated EBML `<element>` which is undefined by this document but MAY be defined by the associated EBML Document Type. The `<extension>` element MUST contain a `type` attribute and also MAY contain any other attribute or sub-element as long as the EBML Schema remains as a well-formed XML Document. All `<extension>` elements MUST be sub-elements of the `<element>`.
+
+### \<extension> Attributes
+
+#### type
+
+The type attribute should reference a name or identifier of the project or authority associated with the contents of the `<extension>` element.
+
+The type attribute is REQUIRED.
 
 ### XML Schema for EBML Schema
 
--- a/EBMLSchema.xsd
+++ b/EBMLSchema.xsd
@@ -28,6 +28,8 @@
         minOccurs="0" maxOccurs="unbounded"/>
       <xs:element name="restriction" type="restrictionType"
         minOccurs="0" maxOccurs="1"/>
+      <xs:element name="extension" type="extensionType"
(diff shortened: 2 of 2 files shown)


specification.markdown
```diff
@@ -312,7 +312,7 @@ The version attribute is REQUIRED within the `<EBMLSchema>` Element.
 
 Each `<element>` defines one EBML Element through the use of several attributes that are defined in [EBML Schema Element Attributes](#ebmlschema-attributes). EBML Schemas MAY contain additional attributes to extend the semantics but MUST NOT conflict with the definitions of the `<element>` attributes defined within this document.
 
-The `<element>` nodes contain a description of the meaning and use of the EBML Element stored within one or more `<documentation>` sub-elements and zero or one `<restriction>` sub-element. All `<element>` nodes MUST be sub-elements of the `<EBMLSchema>`.
+The `<element>` nodes contain a description of the meaning and use of the EBML Element stored within one or more `<documentation>` sub-elements followed by one zero or one `<restriction>` sub-element followed by zero or one `<extension>` sub-element.  All `<element>` nodes MUST be sub-elements of the `<EBMLSchema>`.
```
retokromer: To me, adding a comma before both `followed by` makes the sentence easier to read, but I am not an English speaker.

robUx4:
I'd prefer the last commit (type -> purpose) to be a separate PR as it seems unrelated to the extension PR.

robUx4:
I like this idea a lot.

I think it would be nicer to allow more than one extension and have them name. They could also be extended later. For example the cppname could have extra things between library version, comments to add in the code of a particular library, etc.

Something like this:
```xml
<element name="AttachmentLink" path="0*1(\Segment\Tracks\TrackEntry\AttachmentLink)" id="0x7446" type="uinteger" maxOccurs="1" minver="1" maxver="3" range="not 0">
  <documentation lang="en">The UID of an attachment that is used by this codec.</documentation>
  <extension type="webm" allowed="0"/>
  <extension type="libmatroska" cppname="TrackAttachmentLink">
    <comment>attachment used by the track content<comment>
  </extension>
</element>
```

On the other end from the CELLAR meeting I thought we decide to strip the extra attributes from the main document and only maintain that one. The extra attributes would be put in another document (so could be using this format). So do we want to allow all kinds of "meaningless" (in the EBML way) attributes to a normative document ? Or are we just defining the separate document we might keep on the side ?

dericed:
Based on comments and to clear merge conflicts, this PR is updated. Now the extension element may be repeated, the use of it is fairly ungoverned by EBML. I think the historic details of divx and webm are meaningful in a way to matroska so think it could be interesting to have an unofficial but acceptable way to let them stay in the schema, rather than us working to maintain two documents.

robUx4:
In the case of `cppname` and `divx` they are both legacies. Once we settle a way to use them we will not have to update anything after that. I'd favor a separate file saying which EBML path needs a special `cppname` (which should actually be part of libmatroska, not the Matroska repo) or `divx`, which we saw will probably remain in the spec but forbidden.

For WebM this should also be a separate file that we may or may not maintain. We do not want it to be part of the normative RFC.

robUx4:
So, that being said, it feels like this `extension` element would be OK but, at least for Matroska, not in the main normative document. I can understand there would be some cases where it's useful to people, so it's probably a good addition. But we need to do it right as it will be part of the EBML specifiation and XSD validation.

I'm not sure what `processContents="skip"` in XSD means but it sounds like anything could be put in there. So `<extension><webm>1</webm></extension>` would be OK ? It might cause problem with our XSLT file being fed such files if we don't know anything about the format.

I would prefer something a little more restrictive like `<extension name="webm" value=""/>` plus extra attributes allowed. At least tools that have no idea what's in the extensions could still output something meaningful about it.

robUx4:
Rebased so it can be merged.

dericed:
Yes the `skipContents` means that the validator ignores the contents, as in EBML I'm thinking of leaving it to the DocType about what goes inside. I am ok with `<extension name="webm" value=""/>` style although it is more limited, but could avoid skipContents.

dericed:
@robUx4 I added a @type attribute for the extension element so that it's contextualized in a consistent manner. I preferred @type to @name but don't feel too strongly. Once this is merged, I can adjust the corresponding pr at matroska.

dericed:
rebased and squashed to address merge conflict

PR #271, opened 2019-07-13
add implementation_note element

dericed:
as discussed in this pr https://github.com/cellar-wg/matroska-specification/pull/272#issue-219728738

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -312,7 +312,7 @@ The version attribute is REQUIRED within the `<EBMLSchema>` Element.
 
 Each `<element>` defines one EBML Element through the use of several attributes that are defined in [EBML Schema Element Attributes](#ebmlschema-attributes). EBML Schemas MAY contain additional attributes to extend the semantics but MUST NOT conflict with the definitions of the `<element>` attributes defined within this document.
 
-The `<element>` nodes contain a description of the meaning and use of the EBML Element stored within one or more `<documentation>` sub-elements followed by one zero or one `<restriction>` sub-element. All `<element>` nodes MUST be sub-elements of the `<EBMLSchema>`.
+The `<element>` nodes contain a description of the meaning and use of the EBML Element stored within one or more `<documentation>` sub-elements, optional `<implementation_note>` sub-elements, and zero or one `<restriction>` sub-element. All `<element>` nodes MUST be sub-elements of the `<EBMLSchema>`.
 
 ### \<element> Attributes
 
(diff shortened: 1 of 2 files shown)


specification.markdown
```diff
@@ -476,6 +476,18 @@ A purpose attribute distinguishes the meaning of the documentation. Values for t
...
+
+### \<implementation_note> Attributes
+
+#### note_type
+
+The note_type attribute references which of the `<element>`'s attributes that the implementation_note is in regards to. The note_type attribute MUST be set to one of the following values (corresponding to that attribute of the parent `<element>`): `minOccurs`, `maxOccurs`, `range`, `length`, or `default`. The `<implementation_note>` SHALL supersede the parent `<element>`'s attribute that is named in the `note_type` attribute. An `<element>` SHALL not have more than one `<implementation_note>` of the same `note_type`.
```
retokromer: `SHALL not` > `SHALL NOT`

dericed: updated

robUx4:
Needs an update after 9782f94338596993747dca4f5c6f87e00b0dbf8b (renaming `xsd:` to `xs:`).

specification.markdown
```diff
@@ -476,6 +476,18 @@ A purpose attribute distinguishes the meaning of the documentation. Values for t
...
+
+### \<implementation_note> Attributes
+
+#### note_type
+
+The note_type attribute references which of the `<element>`'s attributes that the implementation_note is in regards to. The note_type attribute MUST be set to one of the following values (corresponding to that attribute of the parent `<element>`): `minOccurs`, `maxOccurs`, `range`, `length`, or `default`. The `<implementation_note>` SHALL supersede the parent `<element>`'s attribute that is named in the `note_type` attribute. An `<element>` SHALL NOT have more than one `<implementation_note>` of the same `note_type`.
```
robUx4: no 'minver' / 'maxver' ? For DivX we MAY explain why elements not found in Matroska can actually be found in Matroska v2 files.

dericed: Good point. Will add.

dericed: added

robUx4:
Rebased so it can be merged (renamed `<xsd:` to `<xs:`).

It will still conflict with #270, whichever is merged first.

dericed:
The prior rebase undid an earlier xsd-cleaning pr, so I cleaned this up and force pushed it. I recommend that this be considered before https://github.com/cellar-wg/ebml-specification/pull/270 as I think there's more consensus here, whereas the extensions may not be necessary to move forth with work on the Matroska EBML Schema.

robUx4:
LGTM

PR #284, opened 2019-09-24
bump to version 12

dericed:


dericed:
to merge, after 11 is in https://github.com/cellar-wg/ebml-specification/releases

PR #285, opened 2019-10-15
this makes some adjustments to maintain the rfc line length requirements

dericed:


specification.markdown
```diff
@@ -359,7 +359,7 @@ The path defines the allowed storage locations of the EBML Element within an EBM
 The path attribute is REQUIRED.
 
 ```
-EBMLFullPath             = EBMLElementOccurrence "(" EBMLMasterPath ")"
```
dericed: @robUx4, here I just removed the min white space needed to adhere to the line length needs of the output, but perhaps it would be better to rename `VariableParentOccurrence` to something shorter and shift everything to the left

dericed:
@robUx4 nudge. Can we merge this before and post an update before the meeting. Advice on renaming VariableParentOccurrence to a shorter phrase?

specification.markdown
robUx4: I'll rename *Variable* to *Global*

dericed:
thx @robUx4

PR #288, opened 2019-10-20
rename note_type to note_atttribute to make it clear we add rule on an attribute

robUx4:
IMO the name is confusing.


dericed:
I added a commit to fix a link to the note_attribute section, LGTM now.

PR #289, opened 2019-10-20
clarify value of default

dericed:
Along with https://github.com/cellar-wg/matroska-specification/pull/272, this should resolve https://github.com/cellar-wg/ebml-specification/issues/224.

In other words, if you have the schema say:

```
<element name="OutputSamplingFrequency" default="SamplingFrequency"/>
```

and the file says
```xml
<SamplingFrequency>44100</SamplingFrequency>
```

then the meaning should be:
```xml
<SamplingFrequency>44100</SamplingFrequency>
<OutputSamplingFrequency>SamplingFrequency</OutputSamplingFrequency>
```

and not
```xml
<SamplingFrequency>44100</SamplingFrequency>
<OutputSamplingFrequency>44100</OutputSamplingFrequency>
```

So the default attribute should store what would be given by interpreting what the EBML Data of the unstored Element should be, and not a reference to another element. However, default values can be made to be associated with a related Element via an implementation_note.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -450,7 +450,9 @@ The length attribute is OPTIONAL. If the length attribute is not present for tha
 
 #### default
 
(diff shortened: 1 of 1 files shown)


dericed:
Ug, I am missing a commit left unpushed on another computer. The missing line was something like:
> An unwritten mandatory Element with a declared default value is semantically equal to if that Element was written with the default value stored as the Element Data.

dericed:
added a commit for review

robUx4:
The `equal to if` wording sound odd. But if that's proper English, LGTM.

dericed:
wording updated

PR #291, opened 2019-10-21
Xml2rfc version 3 updates

dericed:
fixes a few issues with the version 3 upload and from idnits


specification.markdown
```diff
@@ -982,7 +982,7 @@ This document creates a new IANA Registry called "CELLAR EBML Element ID Registr
 
 Element IDs are described in section Element ID. Element IDs are encoded using the VINT mechanism described in section (#variable-size-integer) can be between one and five octets long. Five octet long Element IDs are possible only if declared in the header.
 
-This IANA Registry only applies to Elements that can be contained in the EBML Header, thus including [Global Elements](#global-elements). Elements only found in the EBML Body have their own set of independent Element IDs and are not part of this IANA Registry.
+This IANA Registry only applies to Elements that can be contained in the EBML Header, thus including Global Elements. Elements only found in the EBML Body have their own set of independent Element IDs and are not part of this IANA Registry.
```
robUx4: Is the link gone on purpose ?

dericed: The xml2rfc processor transforms `(#global-elements)` into `Section 1.2.3`, so in these usages it seems better to simply use our defined term and say "thus including Global Elements" rather than "thus including Section 1.2.3". An alternative could be "thus including Global Elements (see Section 1.2.3)".

PR #296, opened 2019-10-24
add example of using a valid master element within an invalid master …

dericed:
This responds to the request for an example at the end of https://mailarchive.ietf.org/arch/msg/cellar/koyh6rLPjpqwmL9iUx2xouKD2g4.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -864,6 +864,25 @@ The following scenarios describe events to consider when reading EBML Documents 
 
 If a Master Element contains a CRC-32 Element that doesn't validate, then the EBML Reader MAY ignore all contained data except for Descendant Elements that contain their own valid CRC-32 Element.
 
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -864,6 +864,25 @@ The following scenarios describe events to consider when reading EBML Documents
...
+```xml
+<CONTACT>
+    <CRC-32>c119a69b</CRC-32><!- does not validate -->
+    <NAME>
+        <CRC-32>1f59ee2b</CRC-32><!- does not validate -->
+        <FIRST-NAME>invalid data</FIRST-NAME>
+        <LAST-NAME>invalid data</LAST-NAME>
+    </NAME>
+    <ADDRESS>
+        <CRC-32>df941cc9</CRC-32><!- validates -->
```
retokromer: All three comments should start with `<!--` not `<!-`.

dericed: updated

robUx4:
LGTM, very good idea

dericed:
I changed the language for clarity a little.

Issue #292, opened 2019-10-22
address conflict between unwritten minOccurs attribute and implementation note


dericed:
After a change in the EBML specification to clarify this, a corresponding change may be needed in the matroska specification.

See discussion on this issue at https://github.com/cellar-wg/matroska-specification/pull/272#issuecomment-545003625

dericed:
This is similar to my earlier use of "see note". I'd suggest to not include brackets within the special keyword.

robUx4:
As with any keyword it must not be mistaken for a value. It could be used as a string, in which case it should be around double quotes (we should mention that in the specs if not already ?). So I guess no brackets would work too. It wouldn't be confused with a string value.

dericed:
The other way to address it would be to adjust this line:
> If the minOccurs attribute is not present then that EBML Element has a minOccurs value of 0.

Perhaps:
> If the Element has an implementation_note with a note_attribute of "minOccurs" than that implementation_note describes the minOccurs and the minOccurs attribute MUST be ignored, else if the minOccurs attribute is not present then that EBML Element has a minOccurs value of 0.

robUx4:
I think it's too specific to how we use it for now. For example an element may be mandatory (minOccurs=1) by default and implementation_note specify when it's not mandatory or when it has to be found twice.

More generally I think the implementation_note should specify each value.

Not sure the value should actually be set with the "regular" value in XML Schema as it can't really be mechanically interpreted. That would be misleading.

robUx4:
(replying to myself) Not setting the `minOccurs` means it uses the default value (0).  Also the path requires an `EBMLMinOccurrence`, if not written it means the value is 0. There is no `<unset>` value.

So either we add the concept of unset value (`-` sign for EBMLMinOccurrence/EBMLMaxOccurrence for example) or we have to pick the best/most common default value and explain in which case the value differs in `implementation_note`.

I think the unset option is better because it doesn't mislead the reader and programs that would parse the path to verify the validity of an element and would have no way of parsing the human-readable text `implementation_note` to tell when the default value is OK or not.

dericed:
I can't remember but is there a reason to have EBMLMinOccurrence/EBMLMaxOccurrence within the path as they are redundant to the minOccurs/maxOccurs attributes?

robUx4:
[...]

PR #297, opened 2019-10-24
Add implementation_note example

dericed:


dericed:
This PR responds to issues noted at https://mailarchive.ietf.org/arch/msg/cellar/xy0WUr07zuMhwotQrybYN18mI30.

specification.markdown
```diff
@@ -535,6 +535,40 @@ An `<element>` SHALL NOT have more than one `<implementation_note>` of the same
 
 The note\_attribute attribute is REQUIRED.
 
+#### \<implementation_note> Example
+
+The following fragment of an EBML Schema demonstrates how an `<implementation_note>` is used. In this case an EBML Schema documents a list of items that are described with an optional cost. The Currency Element uses an `<implementation_note>` to say that the Currency Element is REQUIRED if the Cost Element is set, otherwise not.
+
+```xml
+<element name="Item" path="1*1(\Items\)" id="0x4025" type="master">
```
robUx4: There shouldn't be a `\` at the end of the path.

specification.markdown
```diff
@@ -535,6 +535,40 @@ An `<element>` SHALL NOT have more than one `<implementation_note>` of the same
...
+  <documentation lang="en" purpose="definition">
+    An set of items.
```
robUx4: `A` typo

specification.markdown
robUx4: `minOccurs` 1 is not the default value so should be set in Schema.
`maxOccurs` 1 is not the default value so should be set in Schema.

specification.markdown
```diff
@@ -535,6 +535,40 @@ An `<element>` SHALL NOT have more than one `<implementation_note>` of the same
...
+  </documentation>
+</element>
+<element name="Currency" path="0*1(\Items\Item\Currency)" id="0x403F"
+  type="string" maxOccurs="1">
+  <documentation lang="en" purpose="definition">
+    The currency of the item's cost.
+  </documentation>
+  <implementation_note note_attribute="minOccurs">
+    Currency MUST be set (minOccurs=1) if the associated Item stores
+    a Cost, elseCurrent MAY be unset (minOccurs=0).
```
robUx4: missing space after `else`.

specification.markdown
```diff
@@ -535,6 +535,40 @@ An `<element>` SHALL NOT have more than one `<implementation_note>` of the same
...
+    An item.
+  </documentation>
+</element>
+<element name="Cost" path="0*1(\Items\Item\Cost)" id="0x4024"
+  type="float" minOccurs="0" maxOccurs="1">
+  <documentation lang="en" purpose="definition">
+    The cost of the item, if any.
+  </documentation>
+</element>
+<element name="Currency" path="0*1(\Items\Item\Currency)" id="0x403F"
```
robUx4: `maxOccurs` 1 is not the default value so should be set in Schema.

robUx4: There's a question in #292 whether the minOccurs should be included in the Schema since the value may change depending on the context. Here we see there's another issue, the minOccurs is also included in the path. Here it's the default value, so not writing it also has a different meaning (it's the default).

robUx4: Resolved as it matches the current expected behavior. It may have to be changed once we find a solution for #292

dericed:
Thanks for your updates, will merge.

PR #299, opened 2019-10-27
add basic CI build

robUx4:
It generates the .xml, .txt and .html files

Later we can add verifications on the validity of some files or generate the .xml file to publish when we tag for a release.

**This requires the project to enable "[Github Actions](https://github.com/features/actions)" which is currently in beta.**

You can check the result here: https://github.com/robUx4/ebml-specification/actions
It's a bit long because everytime we have to get mmark and xml2rfc. Maybe it could be better with a Docker, but I've never used that.

Once it's merged I suppose it will be run on Pull Requests as well.

Once it's working I'll do the same on the Matroska specification where we can also validate the XML files.


dericed:
Not sure how this works, but looks very interesting. So github would auto-generate the outputs?

robUx4:
The first goal is to make sure the build actually works (makefile is OK, files to build are OK and don't produce an error).

It's possible to keep the generated files (to upload on IETF) or automatically create a release package (the XML, HTML, TXT files for example) when a version is tagged.

robUx4:
(I'm also totally new to this but wanted too give it a try so anyone can tell ahead if the PR they propose are valid or not)

robUx4:
BTW I called it CI, but I'm not sure it's equivalent to the Continuous Integration proposed by Github. But that's how I found about Github Actions. There are other solutions involving Docker to do the same thing, not always free and not as convenient as this.

robUx4:
Given we haven't enabled Actions yet, it has no effect. I'll keep monitoring in my fork to find out when there are issues.

dericed:
@robUx4, curious if there's any updates on Actions. Is it working in your fork?

robUx4:
Yes it's working https://github.com/robUx4/ebml-specification/actions

Actions is not out of beta and available to anyone. We just have to press the button here https://github.com/cellar-wg/ebml-specification/actions

robUx4:
There might some work need to show the status in a Pull Request and disable merging if something doesn't pass (but we're all responsible enough to avoid that).

PR #302, opened 2019-11-03
Remove occurences from the path

robUx4:
Ref. #292 

The Matroska Schema will need to be updated when this is merged.


robUx4:
I'd like to go ahead with this because it's blocking my "code generators" from the Matroska EBML Schema.

IIRC the goal was originally that with just a path you can tell how to interpret the element. But that's not the case, there's not the type for example. In the end the Schema is more complete.

The path is still important to describe new elements in "foreign" documents without ambiguity.

specification.markdown
```diff
@@ -427,17 +417,30 @@ The id attribute is REQUIRED.
 
 Within an EBML Schema, the XPath of `@minOccurs` attribute is `/EBMLSchema/element/@minOccurs`.
 
-An integer expressing the minimum permitted number of occurrences of this EBML Element within its Parent Element. The minOccurs value MUST be equal to the EBMLMinOccurrence value of the path.
+The minOccurs is an integer representing the minimum permitted number of occurrences of this EBML Element within its Parent Element.
+
+Each instance of the Parent Element MUST contain at least this many instances of this EBML Element.
```
dericed: Perhaps 'contain' is the wrong word. As minOccurs=1 with a default value does not need to be 'contained' within the parent.

specification.markdown
```diff
@@ -427,17 +417,30 @@ The id attribute is REQUIRED.
...
-An integer expressing the maximum permitted number of occurrences of this EBML Element within its Parent Element. The maxOccurs value MUST be equal to the EBMLMaxOccurrence value of the path.
+The maxOccurs is an integer representing the maximum permitted number of occurrences of this EBML Element within its Parent Element.
+
+Each instance of the Parent Element MUST contain at most this many instances of this EBML Element.
+If the EBML Element has an empty EBMLParentPath then maxOccurs refers to constraints on the occurrence of the EBML Element within the EBML Document.
+
+The maxOccurs attribute is OPTIONAL. If the maxOccurs attribute is not present then there is no upper bound for the permitted number of occurrences of this EBML Element within its Parent Element or within the EBML Document depending on whether the EBMLParentPath of the EBML Element is empty or not.
 
-The maxOccurs attribute is OPTIONAL. If the maxOccurs attribute is not present then that EBML Element has no maximum occurrence, similar to unbounded in the XML world.
+The semantic meaning of maxOccurs within an EBML Schema is analogous to the meaning of unbounded within an XML Schema.
```
dericed: This doesn't match line 430 well. I think you mean to say that ebml:minOccurs is like xml:minOccurs but then say ebml:maxOccurs is like xml:unbounded. But it is like xml:unbounded. I suppose an unstated ebml:maxOccurs is like an xml:maxOccurs that is set to "unbounded".

dericed:
After merging https://github.com/cellar-wg/ebml-specification/pull/303 this needs a rebase.

specification.markdown
robUx4: This is the same wording that was used in `EBMLMinOccurrence`, dating back to 40fe4cb33d7e5d06e9f58d389f95d425437071ba.

In the next sentence(s) it actually explains in which case the element can be omitted: `EBML Elements with minOccurs set to "1" that also have a default value declared are not REQUIRED to be stored`.
It was all on the same line before and still in the same paragraph.

For maxOccurs there's no such thing. How about: 
`Each instance of the Parent Element MUST contain at most this many instances of this EBML Element, including the unwritten mandatory element with a default value, see (#note-on-the-use-of-default-attributes-to-define-mandatory-ebml-elements).`

dericed:
Any further comments? I suggest merging soon.

robUx4:
If the `maxOccurs` change is OK with you, you can merge.

PR #303, opened 2019-11-03
Stronger typing of attribute values in the XML Schema

robUx4:


dericed:
Tested again and this validates the current `matroska_xsd.xml`. Thanks for making this more clear.

PR #306, opened 2019-12-10
Response to 2019 12 03 review

dericed:
This almost completely responds to the review at https://datatracker.ietf.org/wg/cellar/documents/#roman-danyliw. cc @mcr 

I didn't respond to 
> Section 11.1.5.3.  Are there any uniqueness properties for an id attribute?  Drawing a parallel from XML,  I would have thought that each EBML element would have unique ID per doctype (say like an xml:id)

EBML doesn't really have the concept of xml:id. ebml:id is like xml:name whereas xml:id is like an anchor for a specific use of an element within a document.


specification.markdown
```diff
@@ -2,9 +2,9 @@
 
 EBML, short for Extensible Binary Meta Language, specifies a binary and octet (byte) aligned format inspired by the principle of XML (a framework for structuring data).
 
-The goal of this document is to define a generic, binary, space-efficient format that can be used to define more complex formats using an EBML Schema. EBML is used by the multimedia container [Matroska](https://github.com/Matroska-Org/matroska-specification/). The applicability of EBML for other use cases is beyond the scope of this document.
+The goal of this document is to define a generic, binary, space-efficient format that can be used to define more complex formats using an EBML Schema. EBML is used by the multimedia container, Matroska [@!Matroska]. The applicability of EBML for other use cases is beyond the scope of this document.
```
robUx4: The `!` makes it normative. I think `[@Matroska]` is the format we want to use here.

specification.markdown
```diff
@@ -2,9 +2,9 @@
...
 
-The definition of the EBML format recognizes the idea behind HTML and XML as a good one: separate structure and semantics allowing the same structural layer to be used with multiple, possibly widely differing semantic layers. Except for the EBML Header and a few Global Elements this specification does not define particular EBML format semantics; however this specification is intended to define how other EBML-based formats can be defined, such as the audio-video container formats Matroska and WebM.
+The definition of the EBML format recognizes the idea behind HTML and XML as a good one: separate structure and semantics allowing the same structural layer to be used with multiple, possibly widely differing semantic layers. Except for the EBML Header and a few Global Elements this specification does not define particular EBML format semantics; however this specification is intended to define how other EBML-based formats can be defined, such as the audio-video container formats Matroska and WebM [@?WebM].
```
robUx4: This also sounds like it's a normative reference. `[@WebM]` looks cleaner. It does show up in the non normative references though.

specification.markdown
```diff
@@ -108,7 +106,7 @@ The VINT\_DATA portion of the Variable Size Integer includes all data that follo
 
 ## VINT Examples
 
-This table shows examples of Variable Size Integers with lengths from 1 to 5 octets. The Usable Bits column refers to the number of bits that can be used in the VINT\_DATA. The Representation column depicts a binary expression of Variable Size Integers where VINT\_WIDTH is depicted by `0`, the VINT\_MARKER as `1`, and the VINT\_DATA as `x`.
+[@tableUsableBits] shows examples of Variable Size Integers with lengths from 1 to 5 octets. The Usable Bits column refers to the number of bits that can be used in the VINT\_DATA. The Representation column depicts a binary expression of Variable Size Integers where VINT\_WIDTH is depicted by `0`, the VINT\_MARKER as `1`, and the VINT\_DATA as `x`.
```
robUx4: Good.

It seems Table 5 and Table 15 are not referenced. Not sure it's important.

robUx4:
The `id` is not unique. The path should be. We may add this in the text.

specification.markdown
dericed: do you suggest captions for them?

specification.markdown
dericed: Thx. I changed the `!` to `?`.

specification.markdown
dericed: `?` means informative but no `?|!` also defaults to informative. I prefer the clarity of saying the symbol.

dericed:
I added https://github.com/cellar-wg/ebml-specification/issues/307 and https://github.com/cellar-wg/ebml-specification/issues/308 to cover remaining issues from the review and this PR. I suggest to merge this.

specification.markdown

specification.markdown
robUx4: I added some.

robUx4:
Oops, didn't see it's merged.

PR #310, opened 2019-12-12
add statement about uniqueness of id and path within an ebml schema

dericed:
fixes https://github.com/cellar-wg/ebml-specification/issues/308

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -417,7 +417,7 @@ Within an EBML Schema, the XPath of `@id` attribute is `/EBMLSchema/element/@id`
 
 The Element ID encoded as a Variable Size Integer expressed in hexadecimal notation prefixed by a 0x that is read and stored in big-endian order. To reduce the risk of false positives while parsing EBML Streams, the Element IDs of the Root Element and Top-Level Elements SHOULD be at least 4 octets in length. Element IDs defined for use at Root Level or directly under the Root Level MAY use shorter octet lengths to facilitate padding and optimize edits to EBML Documents; for instance, the Void Element uses an Element ID with a one octet length to allow its usage in more writing and editing scenarios.
 
-The `@id` value does not need to be unique within an EBML Schema, but the value of the `@path` of that Element MUST be unique within an EBML Schema. In other words, the Element ID of any Element found within an EBML Document MUST only match a single `@path` of its corresponding EBML Schema, but a separate instance of that Element ID value defined by the EBML Schema MAY  occur within a different `@path`.
+The Element ID of any Element found within an EBML Document MUST only match a single `@path` of its corresponding EBML Schema, but a separate instance of that Element ID value defined by the EBML Schema MAY occur within a different `@path`. Two `@path` of two different Element using the same `@id` MUST NOT have a the same EBMLParentPath. Two Elements MUST NOT use the same `@id` if one of their common Parent Element could be an Unknown-Size Element.
 
 The id attribute is REQUIRED.
 
--- a/specification.markdown
+++ b/specification.markdown
@@ -409,7 +409,7 @@ If the path contains an EBMLPathAtomRecursive part, the EBML Element can occur w
 
 As an example, a `path` of `1*(\Segment\Info)` means the element Info is found inside the Segment elements at least once and with no maximum iteration. An element SeekHead with path `0*2(\Segment\SeekHead)` may not be found at all in its Segment parent, once or twice but no more than that.
 
-The `@path` value MUST be unique within the EBML Schema. An `@id` found in the EBML tree of an EBML Document can only match one `@path` unambigously.
+The `@path` value MUST be unique within the EBML Schema. The `@id` corresponding to this `@path` MUST NOT be used in another EBML Element with the same EBMLParentPath as this `@path`.
 
 #### id
 
--- a/specification.markdown
+++ b/specification.markdown
@@ -417,7 +417,7 @@ Within an EBML Schema, the XPath of `@id` attribute is `/EBMLSchema/element/@id`
 
 The Element ID encoded as a Variable Size Integer expressed in hexadecimal notation prefixed by a 0x that is read and stored in big-endian order. To reduce the risk of false positives while parsing EBML Streams, the Element IDs of the Root Element and Top-Level Elements SHOULD be at least 4 octets in length. Element IDs defined for use at Root Level or directly under the Root Level MAY use shorter octet lengths to facilitate padding and optimize edits to EBML Documents; for instance, the Void Element uses an Element ID with a one octet length to allow its usage in more writing and editing scenarios.
 
-The `@id` value does not need to be unique within an EBML Schema, but the value of the `@path` of that Element MUST be unique within an EBML Schema. In other words when an `@id` is found in the tree of an EBML Element, it MUST only match a single `@path`. But the same `@id` MAY be found in a different `@path`.
+The `@id` value does not need to be unique within an EBML Schema, but the value of the `@path` of that Element MUST be unique within an EBML Schema. In other words, the Element ID of any Element found within an EBML Document MUST only match a single `@path` of its corresponding EBML Schema, but a separate instance of that Element ID value defined by the EBML Schema MAY  occur within a different `@path`.
 
 The id attribute is REQUIRED.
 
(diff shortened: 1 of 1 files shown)


robUx4:
I added some text to clarify that. Also the path is not stored in the file so the uniqueness only means something when trying to find the `path` of an `id`

specification.markdown
```diff
@@ -409,12 +409,16 @@ If the path contains an EBMLPathAtomRecursive part, the EBML Element can occur w
 
 As an example, a `path` of `1*(\Segment\Info)` means the element Info is found inside the Segment elements at least once and with no maximum iteration. An element SeekHead with path `0*2(\Segment\SeekHead)` may not be found at all in its Segment parent, once or twice but no more than that.
 
+The path value MUST be unique within the EBML Schema. So an `@id` found in the tree of EBML Elements can only match one path.
```
dericed: I don't understand the last sentence. Is the `tree of EBML Elements` referring to an `EBML Schema` or an `EBML Document`? Also if the `@id` does not need to be unique, then couldn't it match more than one path?

Also I'd suggest to reference 'path' as \`@path\`.

robUx4: > Also I'd suggest to reference 'path' as `@path`.

OK to be coherent with the text around. But we never introducted the notion of @<something> which is an XML/XPath thing, not an EBML Schema thing.

robUx4: The `tree of EBML Elements` refers to the data found in an EBML Document.

The same `id` can be found in many places. The `id` is only used to find the semantic meaning of the data found at a given position in the tree. For example you can find 0xAA at `\root\element\tree\here` and `\root\element\tree\leaf\there`. They could have totally different meanings. only the path guarantees they are uniquely interpreted.

robUx4:
Let me know if the wording is better. (suggestions welcome)

dericed:
We reference XPath and consistently include the @ in the names of the attributes so I think this practice is okay. I am still hesitant to introduce a new concept of `EBML tree` which is used only here and not elsewhere.

How about:
> In other words, the Element ID of any Element found within an EBML Document MUST only match a single `@path` of its corresponding EBML Schema, but a separate instance of that Element ID value MAY defined by the EBML Schema to occur within a different `@path`.

Also I realize that this may cause an issue with finding an unknown sized element's end. For instance if `\root\level1\level2\elt` is unknown sized, but we define two Elements (same id, different path):
- `\root\level1\level2\elt\whereever`
- `\root\level1\level2\whereever`

Then should the occurrence of `whereever` end the element `elt` or not?

robUx4:
I reworked the text to include your change and make the `@id` unique per EBMLParentPath.

robUx4:
There is indeed an issue with finding the end of Unknown Size element. We could say the "closest" EBMLParentPath should be preferred. But some formats may want the other way around.

We could also say that non-unique elements cannot be used to find the boundaries of Unknown Size elements. But then a format may define no unique element and someone add one later, making the initial assumption wrong.

So maybe we should not allow multiple use of the same ID after all.

retokromer:
>So maybe we should not allow multiple use of the same ID after all.

Not allowing makes things easier.

dericed:
I consider the allowance of re-use of an `@id` to be a more helpful feature than unknown sized elements (even in Matroska, unknown sized is only allowed in 2 Elements: Segment and Cluster). For instance if starting the design of Matroska from the beginning again, I would suggest that the *Language Elements use the same @id, although there are many paths. We could add a restriction to say that `An EBML Schema MAY define multiple Elements with the same @id but with distinct @path values only if none of those Elements have a Parent Element that could be an Unknown-Size Element`.

Also the issue I point out is the same for Global Elements as it is for Elements with shared id but different path:

If `\root\level1\level2\elt` is unknown sized, then both:
- `\root\level1\level2\elt\whereever`
- `\root\level1\level2\whereever`

and
- `\root\level1\level2\elt\void`
- `\root\level1\level2\void`

provide the same ambiguity about the end of elt.

robUx4:
> An EBML Schema MAY define multiple Elements with the same @id but with distinct @path values only if none of those Elements have a Parent Element that could be an Unknown-Size Element.

I'll modify the text to use something like that.

robUx4:
The case of Global Elements mixed with Unknown Sized is already covered. When encountering a Global Element within an Unknown Size parent, it never goes up one level. Global Elements are always excluded from the cases where we might go up one level. Hence this example:

> `\root\level1\level2\elt\<global>` | Global Element is valid, it's a child of `elt`
`\root\level1\level2\<global>`     | Global Element cannot be assumed to have this path, while parsing `elt` it can only be a child of `elt`

Maybe the same could be done for Elements with the same `@id`.

With the same example where `\root\level1\level2\elt` is an unknown size element. If `\root\level1\level2\0xAA` and `\root\level1\0xAA` exist and you encounter a `0xAA` id, you can't tell which element you found, unless you forbid going up levels for elements with the same id.

The problem with this is that the logic would change as soon as you add new elements to the Schema. The same file that used to be readable because 0xAA was unique, becomes invalid if you introduce `\root\level1\0xAA` to the Schema.

robUx4:
As for the design choices, I think Unknown Sized elements is a key feature of Matroska/EBML because it makes streaming/low latency possible. The reuse of IDs is just fancy if possible but has no impact in the usage/possibilities. It only allows reusing some one octet IDs multiple times if needed.

For things like Language in Matroska, even if you had the same IDs you still need to map it to code that is only usable and interpreted in a particular path. You can reuse a Language class but still need to interpret it differently. In real life you can already reuse the same Language class even though the IDs are not the same. In libebml for example all integers are using the same class to share some code. That doesn't mean all integers have to have the same ID.

robUx4:
So given https://github.com/cellar-wg/ebml-specification/pull/310#issuecomment-568241654 and https://github.com/cellar-wg/ebml-specification/pull/310#issuecomment-568266521 I think there's little use for ID reusing. Only for formats not using the Unknown Size elements. We may keep the possibility though for other formats. It could be used by others (until they realize they'd like to use the Unknown Size element).

Making all IDs unique in a format will have no impact on Matroska.

Also merging 2 formats (like using 2 XML namespaces in the same file) might also produce duplicate IDs for different pathes. The same ID for the same `EBMLParentPath` is certainly not possible. But in different places it could be possible.

specification.markdown
```diff
@@ -409,12 +409,16 @@ If the path contains an EBMLPathAtomRecursive part, the EBML Element can occur w
 
 As an example, a `path` of `1*(\Segment\Info)` means the element Info is found inside the Segment elements at least once and with no maximum iteration. An element SeekHead with path `0*2(\Segment\SeekHead)` may not be found at all in its Segment parent, once or twice but no more than that.
 
+The `@path` value MUST be unique within the EBML Schema. The `@id` corresponding to this `@path` MUST NOT be used in another EBML Element with the same EBMLParentPath as this `@path`.
```
dericed: > I think 'used' is not clear enough as it could imply no elements with maxoccurs >1. I suggest:  `The @id value corresponding to this @path MUST NOT be defined for use within another EBML Element` ...

specification.markdown
```diff
@@ -409,12 +409,16 @@ If the path contains an EBMLPathAtomRecursive part, the EBML Element can occur w
...
+
 #### id
 
 Within an EBML Schema, the XPath of `@id` attribute is `/EBMLSchema/element/@id`.
 
 The Element ID encoded as a Variable Size Integer expressed in hexadecimal notation prefixed by a 0x that is read and stored in big-endian order. To reduce the risk of false positives while parsing EBML Streams, the Element IDs of the Root Element and Top-Level Elements SHOULD be at least 4 octets in length. Element IDs defined for use at Root Level or directly under the Root Level MAY use shorter octet lengths to facilitate padding and optimize edits to EBML Documents; for instance, the Void Element uses an Element ID with a one octet length to allow its usage in more writing and editing scenarios.
 
+The Element ID of any Element found within an EBML Document MUST only match a single `@path` of its corresponding EBML Schema, but a separate instance of that Element ID value defined by the EBML Schema MAY occur within a different `@path`. Two `@path` of two different Element using the same `@id` MUST NOT have a the same EBMLParentPath. Two Elements MUST NOT use the same `@id` if one of their common Parent Element could be an Unknown-Size Element.
```
dericed: I suggest: `Two @path values of two different Elements...` and plural for Parent Element.

specification.markdown
dericed: I responded to my own comment with https://github.com/cellar-wg/ebml-specification/pull/310/commits/629ff4b8dbad177c131aa94f6b55c00a43c32696

specification.markdown
dericed: I responded to my own comment with https://github.com/cellar-wg/ebml-specification/pull/310/commits/629ff4b8dbad177c131aa94f6b55c00a43c32696

dericed:
I added another commit to address the rewording suggestions that I had. The PR now LGTM, but I can't approve my own PR, even though much of it is now written by @robUx4 :)

PR #311, opened 2019-12-15
don't make EBML too generic

### LINKED ISSUE #304: To generic or not to generic
As per [Robert Sparks review](https://mailarchive.ietf.org/arch/msg/cellar/mlXwL0BuYcw6SlHWdlSSvBcvaLU), we may or may not make EBML a generic format usable by others than Matroska. If so we don't even need a IANA registry.

We need to take a definitive choice. If we go the generic way there might need more scrutiny from further people before we can publish. If not it doesn't mean that noone can use EBML for other things (and some already do).

I don't have a strong opinion on either way.

robUx4:
Its main goal is audio/video storage

Fixes #304 

I did not change the text in the main document which seems to be OK as we already state it's for audio/video:

```The goal of this document is to define a generic, binary, space-efficient format that can be used to define more complex formats using an EBML Schema. EBML is used by the multimedia container, Matroska [Matroska]. The applicability of EBML for other use cases is beyond the scope of this document.```

### DIFF
--- a/rfc_frontmatter.markdown
+++ b/rfc_frontmatter.markdown
@@ -36,6 +36,6 @@ fullname="Moritz Bunkus"
 
 .# Abstract
 
-This document defines the Extensible Binary Meta Language (EBML) format as a generalized file format for any type of data in a hierarchical form. EBML is designed as a binary equivalent to XML and uses a storage-efficient approach to build nested Elements with identifiers, lengths, and values. Similar to how an XML Schema defines the structure and semantics of an XML Document, this document defines how EBML Schemas are created to convey the semantics of an EBML Document.
+This document defines the Extensible Binary Meta Language (EBML) format as a general purpose audio/video container format. EBML is designed as a binary equivalent to XML and uses a storage-efficient approach to build nested Elements with identifiers, lengths, and values. Similar to how an XML Schema defines the structure and semantics of an XML Document, this document defines how EBML Schemas are created to convey the semantics of an EBML Document.
 
 {mainmatter}
--- a/rfc_frontmatter.markdown
+++ b/rfc_frontmatter.markdown
@@ -36,6 +36,6 @@ fullname="Moritz Bunkus"
 
 .# Abstract
 
-This document defines the Extensible Binary Meta Language (EBML) format as a general purpose audio/video container format. EBML is designed as a binary equivalent to XML and uses a storage-efficient approach to build nested Elements with identifiers, lengths, and values. Similar to how an XML Schema defines the structure and semantics of an XML Document, this document defines how EBML Schemas are created to convey the semantics of an EBML Document.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: rfc_frontmatter.markdown [hunk 1]
```diff
@@ -36,6 +36,6 @@ fullname="Moritz Bunkus"
 
 .# Abstract
 
-This document defines the Extensible Binary Meta Language (EBML) format as a generalized file format for any type of data in a hierarchical form. EBML is designed as a binary equivalent to XML and uses a storage-efficient approach to build nested Elements with identifiers, lengths, and values. Similar to how an XML Schema defines the structure and semantics of an XML Document, this document defines how EBML Schemas are created to convey the semantics of an EBML Document.
+This document defines the Extensible Binary Meta Language (EBML) format as a general purpose audio/video container format. EBML is designed as a binary equivalent to XML and uses a storage-efficient approach to build nested Elements with identifiers, lengths, and values. Similar to how an XML Schema defines the structure and semantics of an XML Document, this document defines how EBML Schemas are created to convey the semantics of an EBML Document.
```
dericed: Would an `.mks` file of timed text still meet this definition? Also I know @JeromeMartinez is using a 'rawcooked' EBML Document to store DPX headers, but the data is not audio/visual. I suppose I think that EBML is useful for generic data, so not convinced of the need to restrict the implementation to the type of data contained.

robUx4: We could says "general purpose container format designed for audio/video" ?
It tells the focus without restricting to other usage. 

`.mks` are subtitles, which can be considered as video, it's rendered into a display from a text based codec.
A Chapter only file (referencing other segments) doesn't contain any audio/video, it's still meant to be played as audio/video.
My understanding of the RAWcooked data is that it's still tied to the audio/video data, plus it's still very much targeted as audio/video storage.

retokromer: I would also suggest to refrain from limit EBML to audio/video container format.

robUx4: From Adam's Roach review: `The Abstract still claims the much broader scope -- please update it to match the reduced scope in the Introduction.`

The fact we could say it's designed for audio/video wouldn't restrict it to other usage. Also even if we keep the change of this PR, that doesn't mean noone can use the format for other things.

Maybe it would restrict how `DocType` are allowed in the IANA Registry. @mcr  do you know?

dericed: I'd be okay with that. I just want to avoid a consideration that an EBML Document is invalid, because it does not contain audiovisual data (such as mks or rawcooked). And that this statement would not be blocking the registration of a Document Type for non-audiovisual use.

robUx4: Rewrote the sentence so it feels less generic and mention the audio/video is for the design. It doesn't claim to be very generic, nor does it block its usage as a generic format IMO.

robUx4:
ping @adamroach if the text change seems OK.

PR #312, opened 2019-12-16
bump version to 16

dericed:


dericed:
merging as version 15 is posted at https://datatracker.ietf.org/doc/draft-ietf-cellar-ebml/15

PR #314, opened 2019-12-20
drop cellar from iana registry names

dericed:
as suggested at https://mailarchive.ietf.org/arch/msg/cellar/8KYRnY0zqTEpTdr3GIfSWvVVGgo
ping to @mcr as I believe you originally named these registries


dericed:
will merge, @mcr, please comment is you suggest otherwise

PR #316, opened 2019-12-20
Response to barry leiba's review

dericed:
see https://mailarchive.ietf.org/arch/msg/cellar/epuOIwz8VXiu3V9-AbHNa7BRZjY

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -92,9 +92,9 @@ The Element ID and Element Data Size are both encoded as a Variable Size Integer
 
 ## VINT_WIDTH
 
-Each Variable Size Integer begins with a VINT\_WIDTH which consists of zero or many zero-value bits. The count of consecutive zero-values of the VINT\_WIDTH plus one equals the length in octets of the Variable Size Integer. For example, a Variable Size Integer that starts with a VINT\_WIDTH which contains zero consecutive zero-value bits is one octet in length and a Variable Size Integer that starts with one consecutive zero-value bit is two octets in length. The VINT\_WIDTH MUST only contain zero-value bits or be empty.
+Each Variable Size Integer starts with a VINT\_WIDTH followed by a VINT\_MARKER. VINT\_WIDTH is a sequence of zero or more bits of value `0`, and is terminated by the VINT\_MARKER, which is a single bit of value `1`. The total length in bits of both VINT\_WIDTH and VINT\_MARKER is the total length in octets in of the Variable Size Integer.
 
-Within the EBML Header the VINT\_WIDTH of a VINT MUST NOT exceed three bits in length (meaning that the Variable Size Integer MUST NOT exceed four octets in length) except if said VINT is used to express the Element Data Size of an EBML Element with Element Name EBML and Element ID `0x1A45DFA3` (see (#ebml-element)) in which case the VINT\_WIDTH MUST NOT exceed seven bits in length. Within the EBML Body, when a VINT is used to express an Element ID, the maximum length allowed for the VINT\_WIDTH is one less than the value set in the EBMLMaxIDLength Element. Within the EBML Body, when a VINT is used to express an Element Data Size, the maximum length allowed for the VINT\_WIDTH is one less than the value set in the EBMLMaxSizeLength Element.
+The single bit `1` starts a Variable Size Integer with a length of one octet. The sequence of bits `01` starts a Variable Size Integer with a length of two octets. `001` starts a Variable Size Integer with a length of three octets, and so on, with each additional 0-bit adding one octet to the length of the Variable Size Integer.
 
 ## VINT_MARKER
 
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -117,15 +117,15 @@ Octet Length | Usable Bits | Representation
...
-10           | 2            | 0100 0000 0000 0010
-10           | 3            | 0010 0000 0000 0000 0000 0010
-10           | 4            | 0001 0000 0000 0000 0000 0000 0000 0010
-Table: VINT examples depicting the same integer value rendered at different VINT sizes {#tableVariousSizes}
+Integer | Octet Length | As Represented in VINT (binary)         | As Represented in VINT (hexadecimal)
+--------|--------------|----------------------------------------:|------------------------------------:
+2       | 1            |                               1000 0010 |       0x82
+2       | 2            |                     0100 0000 0000 0010 |     0x4002
+2       | 3            |           0010 0000 0000 0000 0000 0010 |   0x200002
+2       | 4            | 0001 0000 0000 0000 0000 0000 0000 0010 | 0x10000002
```
retokromer: I suggest to use the same syntax for both columns: binary without `0b` and hexadecimal without `0x`.

robUx4: I prefer the way it is now. No need to even check the column to understand what's going on.

specification.markdown
```diff
@@ -92,9 +92,9 @@ The Element ID and Element Data Size are both encoded as a Variable Size Integer
 
 ## VINT_WIDTH
 
-Each Variable Size Integer begins with a VINT\_WIDTH which consists of zero or many zero-value bits. The count of consecutive zero-values of the VINT\_WIDTH plus one equals the length in octets of the Variable Size Integer. For example, a Variable Size Integer that starts with a VINT\_WIDTH which contains zero consecutive zero-value bits is one octet in length and a Variable Size Integer that starts with one consecutive zero-value bit is two octets in length. The VINT\_WIDTH MUST only contain zero-value bits or be empty.
+Each Variable Size Integer starts with a VINT\_WIDTH followed by a VINT\_MARKER. VINT\_WIDTH is a sequence of zero or more bits of value `0`, and is terminated by the VINT\_MARKER, which is a single bit of value `1`. The total number of bits (VINT\_WIDTH and VINT\_MARKER combined) is the number of octets in length of the Variable Size Integer.
```
robUx4: `the sum of VINT\_WIDTH and VINT\_MARKER` sounds better than `combined` which is vague mathematically. The total number in question is then more obvious.

specification.markdown
```diff
@@ -92,9 +92,9 @@ The Element ID and Element Data Size are both encoded as a Variable Size Integer
...
 
-Within the EBML Header the VINT\_WIDTH of a VINT MUST NOT exceed three bits in length (meaning that the Variable Size Integer MUST NOT exceed four octets in length) except if said VINT is used to express the Element Data Size of an EBML Element with Element Name EBML and Element ID `0x1A45DFA3` (see (#ebml-element)) in which case the VINT\_WIDTH MUST NOT exceed seven bits in length. Within the EBML Body, when a VINT is used to express an Element ID, the maximum length allowed for the VINT\_WIDTH is one less than the value set in the EBMLMaxIDLength Element. Within the EBML Body, when a VINT is used to express an Element Data Size, the maximum length allowed for the VINT\_WIDTH is one less than the value set in the EBMLMaxSizeLength Element.
+Thus, the single bit `1` starts a Variable Size Integer with a length of one octet.  The sequence of bits `01` starts a Variable Size Integer with a length of two octets. `001` starts a Variable Size Integer with a length of three octets, and so on, with each additional 0-bit adding one octet to the length of the Variable Size Integer.
```
robUx4: `A single bit 1` otherwise it sounds like you're referring to the `1` delimiter above.

specification.markdown
```diff
@@ -283,10 +283,14 @@ The EBML Header documents the EBML Schema (also known as the EBML DocType) that
...
 
+Elements within an EBML Header can be at most 4 octets long, except for the EBML Element with Element Name EBML and Element ID `0x1A45DFA3` (see (#ebml-element)), which can be up to 8 octets long.
+
 ## EBML Body
 
 All data of an EBML Document following the EBML Header is the EBML Body. The end of the EBML Body, as well as the end of the EBML Document that contains the EBML Body, is reached at whichever comes first: the beginning of a new EBML Header at the Root Level or the end of the file. This document defines precisely which EBML Elements are to be used within the EBML Header, but does not name or define which EBML Elements are to be used within the EBML Body. The definition of which EBML Elements are to be used within the EBML Body is defined by an EBML Schema.
 
+Within the EBML Body, the maximum octet length allowed for any Element ID is set by the EBMLMaxIDLength Element and the maximum octet length allowed for any Element Data Size is set by the EBMLMaxSizeLength Element.
```
robUx4: Maybe add the `EBMLMaxIDLength Element of the EBML Header` (same for EBMLMaxSizeLength). To avoid any confusion.

specification.markdown
dericed: I agree with @robUx4. Through the document binary is used with a reference to it being binary, while hexadecimal expressions use a prefix. I prefer the way it is now.

specification.markdown
dericed: the sum of VINT_WIDTH and VINT_MARKER always equals `1` ;)

I changed to:
> The total length in bits of both VINT\_WIDTH and VINT\_MARKER is the total length in octets in of the Variable Size Integer.

specification.markdown
dericed: to be clear, you suggest changing `Thus, the single bit 1` to `Thus, a single bit 1`?

dericed:
updated and rebased in response to @robUx4's comments. Please re-review.

specification.markdown
robUx4: Yes. Or you can drop the `Thus` so it doesn't tie the single bit to the previous paragraph.

dericed: Ok, changed to `The single bit 1`

PR #319, opened 2019-12-24
reword invalid Global Element in unknown length element table

robUx4:


specification.markdown
```diff
@@ -187,7 +187,7 @@ EBML Path of new element           | Status
 `\root\level1\level2\elt`          | Ends the Unknown-Sized Element, as they share the same parent
 `\root\level1\level2\elt\inside`   | Doesn't end the Unknown-Sized Element, it's a child of `elt`
 `\root\level1\level2\elt\<global>` | Global Element is valid, it's a child of `elt`
-`\root\level1\level2\<global>`     | Global Element cannot be assumed to have this path, while parsing `elt` it can only be a child of `elt`
+`\root\level1\level2\<global>`     | Global Element cannot be interpreted with this path, while parsing `elt` a Global Element can only be a child of `elt`
```
kaduk: nit: I think we could change (most of?) these commas to semicolons, but feel free to leave that to the RFC Editor's application of the style book.

robUx4: Done

PR #320, opened 2019-12-24
UTF-8 strings with NULL must still be valid UTF-8

robUx4:

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -1010,7 +1010,7 @@ If a Master Element contains more occurrences of a Child Element than permitted 
 
 Null Octets, which are octets with all bits set to zero, MAY follow the value of a String Element or UTF-8 Element to serve as a terminator.
 An EBML Writer MAY terminate a String Element or UTF-8 Element with Null Octets in order to overwrite a stored value with a new value of lesser length while maintaining the same Element Data Size (this can prevent the need to rewrite large portions of an EBML Document); otherwise the use of Null Octets within a String Element or UTF-8 Element is NOT RECOMMENDED.
-An UTF-8 string with Null octets at the end MUST be a valid UTF-8 string. In other words a Null octet cannot be found inside a multi-octet UTF-8 codepoint.
+A UTF-8 string with Null octets at the end MUST be a valid UTF-8 string. In other words a Null octet cannot be found inside a multi-octet UTF-8 codepoint.
 A string value and a copy of that string value terminated by one or more Null Octets are semantically equal.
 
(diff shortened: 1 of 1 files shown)


robUx4:
That's why I mention "multi-octet UTF-8 codepoint". I don't think you can have 0x80 0x00 as valid UTF-8. ie 0x00 cannot be found in a value UTF-8 codepoint.

specification.markdown
```diff
@@ -1008,7 +1008,11 @@ If a Master Element contains more occurrences of a Child Element than permitted
...
+Null Octets, which are octets with all bits set to zero, MAY follow the value of a String Element or UTF-8 Element to serve as a terminator.
+An EBML Writer MAY terminate a String Element or UTF-8 Element with Null Octets in order to overwrite a stored value with a new value of lesser length while maintaining the same Element Data Size (this can prevent the need to rewrite large portions of an EBML Document); otherwise the use of Null Octets within a String Element or UTF-8 Element is NOT RECOMMENDED.
+An UTF-8 string with Null octets at the end MUST be a valid UTF-8 strings. In other words a Null octet cannot be found inside a multi-octet UTF-8 codepoint.
+An EBML Reader MUST consider the value of the String Element or UTF-8 Element to be terminated upon the first read Null Octet and MUST ignore any data following the first Null Octet within that Element.
```
kaduk: I think this means that if the reader does encounter an invalid encoding such as 0x80 0x00, it will not try to assign meaning to the "incomplete codepoint" and terminate the string after the previous fully-encoded codepoint.  It's unclear whether adding or changing the text would help clarify that or not; the previous sentence takes care of the main concern.

robUx4: Indeed, it's redundant now that we define `An UTF-8 string with Null octets at the end`. It's one or more octets. I'll remove it.

dericed:
I fixed an article in here, but still feel unsure about the phrase: `A UTF-8 string with Null octets at the end MUST be a valid UTF-8 string.` The phrase 'with Null octets at the end' makes it seem like the null octets are part fo the UTF-8 string that is supposed to be valid.

robUx4:
How about `followed by Null octets at the end` ? (at the end is a bit redundant)

dericed:
I suppose I'm uncertain there's a need for the statement at all. And the first and second sentence are written as if they are related but say different things.

How about?
> The Element Data of a UTF-8 Element MUST be a valid UTF-8 string up to whichever comes first: the end of the Element or the first occurring Null octet. The Null octet itself and any following data within that Element SHOULD be ignored.

robUx4:
OK, sounds good to me.

dericed:
updated as suggested with some minor rewording. PR LGTM now.

PR #322, opened 2019-12-24
target a specific version of EBML Header in the Schema

robUx4:


robUx4:
Excellent fix, thanks!

PR #324, opened 2019-12-24
remove remaining text when the path include the occurences

robUx4:

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -735,7 +735,7 @@ This document contains definitions of all EBML Elements of the EBML Header.
 
 name: EBML
 
-path: `1*1(\EBML)`
+path: `\EBML`
 
 id: 0x1A45DFA3
 
@@ -751,7 +751,7 @@ description: Set the EBML characteristics of the data to follow. Each EBML Docum
 
 name: EBMLVersion
 
-path: `1*1(\EBML\EBMLVersion)`
+path: `\EBML\EBMLVersion`
 
 id 0x4286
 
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -737,7 +735,7 @@ This document contains definitions of all EBML Elements of the EBML Header.
 
 name: EBML
 
-path: `1*1(\EBML)`
+path: `\EBML`
```
dericed: I don't understand this change and the subsequent ones. If the PathMinOccurrence is not present then it is understood to be 0, so this edit seems to mean that the EBML Header is no longer required.

dericed (commented):
This commit seems to remove most uses of PathMinOccurrence and PathMaxOccurrence but I don't understand what replaces that. Also if such edits to the `Items` xml is occurring, then the same edits should effect the ebml_schema_example.xml file.

specification.markdown
robUx4: `PathMinOccurrence` and `PathMaxOccurrence` are only use for global elements which `\EBML` and the following ones are not:
> PathMinOccurrence represents the minimum number of element path required between the EBMLFixedParent and the Global Element EBMLElementPath.
> PathMaxOccurrence represents the maximum number of element path possible between the EBMLFixedParent and the Global Element EBMLElementPath

I have to rework the global elements path to make this clearer (#326)

Basically, most pathes don't have occurences in them anymore. Only the Global Elements do (for now).

dericed:
I suggest closing in favor of https://github.com/cellar-wg/ebml-specification/pull/331

robUx4:
No objection although one can be done after the other (there won't be any conflict)

dericed:
closing in favor of the work completed in https://github.com/cellar-wg/ebml-specification/pull/331

PR #330, opened 2019-12-30
add the missing recursive attribute in EBMLSchema.xsd

robUx4:
The attribute is defined in the specification for global elements.


retokromer:
good catch

PR #329, opened 2019-12-30
the EBMLParentPath may only contain the EBMLLastParent

robUx4:
Since they can't be empty and include a path that excluded \pathA\elt

PR #331, opened 2019-12-30
ABNF cleanup

### LINKED ISSUE #326: CRC32/Void pathes don't match the ABNF path rules
They are missing parenthesis. 

The ABNF can probably be simplified as well for global elements.

Fix after #324 is merged.

robUx4:
This PR sits on top of #324 and #329

Follow each commit for more details on the progression of changes that lead to the removal of some of the elements in the ABNF.

The ABNF has also been reordered so that each item is defined after all the elements using it are defined.

Fixes #326


robUx4:
After this Matroska will have pathes recursive pathes like this:

```
\Segment\Chapters\EditionEntry(\ChapterAtom)
\Segment\Chapters\EditionEntry(\ChapterAtom)\ChapterUID

\Segment\Tags\Tag(\SimpleTag)
\Segment\Tags\Tag(\SimpleTag)\TagName
```

There is no min or max level involved.

specification.markdown
```diff
@@ -386,14 +386,13 @@ The path defines the allowed storage locations of the EBML Element within an EBM
 The path attribute is REQUIRED.
 
 ```
-EBMLFullPath          = [EBMLParentPath] EBMLElementPath
-EBMLParentPath        = EBMLFixedParent EBMLLastParent
-EBMLFixedParent       = *(EBMLPathAtom)
+EBMLFullPath          = [EBMLParentPath] [EBMLGlobalParent] 
+                        EBMLElementPath
```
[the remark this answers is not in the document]
robUx4: You understand correctly.

specification.markdown
```diff
@@ -386,14 +386,13 @@ The path defines the allowed storage locations of the EBML Element within an EBM
...
+EBMLParentPath        = 1*(EBMLElementPath)
 EBMLElementPath       = EBMLPathAtom / EBMLPathAtomRecursive
+EBMLPathAtomRecursive = "(" EBMLPathAtom ")"
 EBMLPathAtom          = PathDelimiter EBMLAtomName
-EBMLPathAtomRecursive = "(1*(" EBMLPathAtom "))"
-EBMLLastParent        = EBMLPathAtom / EBMLGlobalParent
-EBMLGlobalParent      = "(" GlobalParentOccurence "\)"
+EBMLGlobalParent      = "(" GlobalParentOccurence ")"
```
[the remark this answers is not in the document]
robUx4: I would prefer not to. Because changing this sentence would be tricky:

> The GlobalParentOccurence part is interpreted as an ABNF Variable Repetition

While `x*y` is a valid repetition in ABNF, `"("x*y")"` or `(x*y)` are not.
Many of the items are split so we can define them in the text.

specification.markdown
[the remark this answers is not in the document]
dericed: OK, so IIUC an EBMLFullPath can be an EBMLParentPath followed by an EBMLElementPath with no EBMLGlobalParent, but in that case how could the boundary between the EBMLParentPath and the EBMLElementPath be determined?

Or is EBMLParentPath only relevant to elements with an EBMLGlobalParent? If so, then  EBMLParentPath could be added to EBMLGlobalParent rather than a component of EBMLFullPath.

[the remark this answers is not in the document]
robUx4: EBMLParentPath is empty or multiple `\abcdefg` parts
EBMLElementPath is non-empty and a `\abcdefg` part.
The EBMLElementPath is the last `\abcdefg` of a full path. The rest is the EBMLParentPath. That's how it is deterministic.

The EBMLGlobalParent is always in the form `(x*y)` which is distinct enough from a EBMLPathAtomRecursive which is in the form `(\abcdef)`.

Void/CRC-32 still include the `\` in the EBMLGlobalParent part. I think it makes it easier to read that it's the repetition of some path, so I'll add it back in EBMLGlobalParent.

robUx4:
These new commits add more cleaning and a differentiation of the last PathAtom so it can be referenced explicitly in the text.

robUx4:
And also used `GlobalPlaceholder` as it's not a real path like the other parts. Just a virtual placeholder for any PathAtom

robUx4:
I added an extra commit in https://github.com/cellar-wg/ebml-specification/tree/global-in-children to add the GlobalPlaceholder in children of a Global Element.

specification.markdown
```diff
@@ -386,14 +386,14 @@ The path defines the allowed storage locations of the EBML Element within an EBM
...
+EBMLParentPath        = 1*(IntermediatePathAtom)
+IntermediatePathAtom  = EBMLPathAtom / EBMLPathAtomRecursive
+EBMLPathAtomRecursive = "(" EBMLPathAtom ")"
 EBMLPathAtom          = PathDelimiter EBMLAtomName
-EBMLPathAtomRecursive = "(1*(" EBMLPathAtom "))"
-EBMLLastParent        = EBMLPathAtom / EBMLGlobalParent
-EBMLGlobalParent      = "(" GlobalParentOccurence "\)"
+GlobalPlaceholder     = "(" GlobalParentOccurence "\)"
 EBMLAtomName          = 1*(EBMLNameChar)
 EBMLNameChar          = ALPHA / DIGIT / "-" / "."
```
dericed: EBMLNameChar doesn't match the spec. I think it should be something like
`EBMLNameChar = ALPHA / DIGIT 0*[ALPHA / DIGIT / "-" / "."]`

specification.markdown
```diff
@@ -404,23 +404,25 @@ PathMaxOccurrence     = 1*DIGIT ; no upper limit
 
 The `*`, `(` and `)` symbols are interpreted as defined in [@!RFC5234].
 
-The EBMLPathAtom part of the EBMLElementPath MUST be equal to the name attribute of the EBML Schema.
+The EBMLAtomName part of the EBMLLastPath MUST be equal to the `@name` attribute of the EBML Schema.
 
 The starting PathDelimiter of the path corresponds to the root of the EBML Document.
 
-In some cases the EBMLLastParent part of the path is an EBMLGlobalParent. A path with a EBMLGlobalParent defines a Global Element; see (#global-elements). Any path that starts with the EBMLFixedParent of the Global Element and matches the occurrences found in the GlobalParentOccurence is a valid path for the Global Element. 
+If the EBMLLastPath is an EBMLPathAtomRecursive, the EBML Element can occur within itself recursively (see (#recursive)).
```
dericed: I'd suggest that this sentence should immediately follow the one on line 407, such as

>The EBMLAtomName part of the EBMLLastPath MUST be equal to the `@name` attribute of the EBML Schema. If the EBMLLastPath is an EBMLPathAtomRecursive, the EBML Element can occur within itself recursively (see (#recursive)).

specification.markdown
robUx4: Actually the `EBMLAtomName` start with be a `ALPHA / DIGIT`. `EBMLNameChar` is the list of possible chars after the first one. But good catch.

specification.markdown

robUx4:
This Matroska branch reflects the changes in this branch.
https://github.com/cellar-wg/matroska-specification/tree/del-path-occurences

Once this branch is merged, that branch should be merged too.

specification.markdown
```diff
@@ -404,23 +404,24 @@ PathMaxOccurrence     = 1*DIGIT ; no upper limit
...
 
-If the path contains an EBMLPathAtomRecursive part, the EBML Element can occur within itself recursively (see (#recursive)).
+The GlobalParentOccurence part is interpreted as an ABNF Variable Repetition.
+The repetition amounts correspond to the amount of unspecified valid Elements that can be found in place of the GlobalPlaceholder for the path to be valid.
 
-As an example, a `path` of `1*(\Segment\Info)` means the element Info is found inside the Segment elements at least once and with no maximum iteration. An element SeekHead with path `0*2(\Segment\SeekHead)` may not be found at all in its Segment parent, once or twice but no more than that.
+PathMinOccurrence represents the minimum number of element path required between the EBMLParentPath and the EBMLLastPath. For example 0 means the EBMLLastPath can be right after the EBMLParentPath, 1 means there has to be at least an element between the EBMLParentPath and the EBMLLastPath. If PathMinOccurrence is not present then that GlobalPlaceholder has a PathMinOccurrence value of 0.
```
dericed: `the minimum number of element path` could be better worded. It feels unclear to me. 

Actually the first sentence is hard to process. For instance with `/Root/(2*)/global` if PathMinOccurrence=2 and there needs to be at least 2 `number of element path required between the EBMLParentPath and the EBMLLastPath`. Does that mean?

```
/Root
/Root/ChildA
/Root/ChildB
/Root/global
```

or
```
/Root/ChildA/ChildB/global
```

Under different interpretations ChildA and ChildB are `between` Root and global.

robUx4: That's `\Root\ChildA\ChildB\global`. I'll see if I can find something bettter.

robUx4: Added some more detailed example, hopefully not too verbose.

robUx4:
Side note, we define `PathMaxOccurrence` but neither in EBML or Matroska it has a different value than the default one. We could get rid of it. Not sure there are cases where semi-global elements make sense.

specification.markdown
```diff
@@ -404,23 +404,36 @@ PathMaxOccurrence     = 1*DIGIT ; no upper limit
...
+The GlobalParentOccurence part is interpreted as an ABNF Variable Repetition.
+The repetition amounts correspond to the amount of unspecified valid Elements that can be found in place of the GlobalPlaceholder for the path to be valid.
 
-PathMaxOccurrence represents the maximum number of element path possible between the EBMLFixedParent and the Global Element EBMLElementPath. It cannot have the value 0 as it would be the Global Element can only be found right after the EBMLFixedParent, in which case it's not a Global Element anymore. If PathMaxOccurrence is not present then there is no upper bound for the permitted number of occurrences of element path possible between the EBMLFixedParent and the Global Element EBMLElementPath.
+PathMinOccurrence represents the minimum number of path elements required between the EBMLParentPath and the EBMLLastPath.
```
dericed: But what does this mean if there is no EBMLParentPath (it is optional)?

specification.markdown
```diff
@@ -404,23 +404,36 @@ PathMaxOccurrence     = 1*DIGIT ; no upper limit
...
 
-As an example, a `path` of `1*(\Segment\Info)` means the element Info is found inside the Segment elements at least once and with no maximum iteration. An element SeekHead with path `0*2(\Segment\SeekHead)` may not be found at all in its Segment parent, once or twice but no more than that.
+PathMaxOccurrence represents the maximum number of path elements possible between the EBMLParentPath and the EBMLLastPath.
+If PathMaxOccurrence is not present then there is no upper bound for the permitted number of occurrences of path elements possible between the EBMLParentPath and the EBMLLastPath.
+PathMaxOccurrence cannot have the value 0 as it would mean the EBMLLastPath can only be found right after the EBMLParentPath, in which case it's not a Global Element anymore and a GlobalPlaceholder MUST NOT be used. 
+
+For example consider an EBML Path `\a(0*1\)\global`, there has to be at most one path element between the EBMLParentPath `\a` and the EBMLLastPath `\global`.
+So the `global` EBML Element can be found inside the `\a` EBML Element or inside the `\a\b` EBML Element.
```
dericed: Like the comment above, I'm not sure how to understand this if it was `(0*1\)\global`.

specification.markdown
robUx4: Yes it's optional. And in the only cases it's used CRC-32 and Void it is empty. In that case it's the root.
Maybe we can mention that an empty `EBMLParentPath` means it's the root.

Alternatively we can rework the ABNF so that the root is always present. In that case the EBMLLastPath would not contain the `\` anymore.

dericed: Hi @robUx4, I tried with the latter option and making EBMLParentPath mandatory, where that can be either `\` as Root or the series of intermediate paths. Please review.

specification.markdown
```diff
@@ -968,7 +968,7 @@ A Global Element can be found in many Parent Elements, allowing the same number
 
 name: CRC-32
 
-path: `(1*\)\CRC-32`
+path: `\(1*\)\CRC-32`
```
robUx4: The idea is that the parent `\` is not part of the element name anymore, ie `\(1*\)CRC-32` or `\(*\)Void`. So if you read it like a pattern for CRC-32 you need at least one extra \ in the path. For Void it can have 0 (or plenty) extra \ in the path.

robUx4:
I reworked the notation for recursive elements (just a +, similar to what was originally used in the Matroska specs).

I also removed the \ from the global element name after the `(*\)` so the number of \ corresponds to the amount of levels you can find the element in. (not referred to as levels)

I redid the occurence test as the ABNF analogy was causing more problem than anything. It's not a repetition as found in ABNF but a range of possible occurences, of elements replacing the global placeholder.

I added a text file with some path examples to show the various possibilities and then ABNF parts they are decomposed into. That really helps finding valid/invalid syntaxes (or reducing the recursive to just a plus).

specification.markdown
```diff
@@ -386,41 +386,60 @@ The path defines the allowed storage locations of the EBML Element within an EBM
...
+EBMLPathAtom          = [IsRecursive] EBMLAtomName PathDelimiter
+EBMLElement           = [IsRecursive] EBMLAtomName
+
 PathDelimiter         = "\"
-GlobalParentOccurence = [PathMinOccurrence] "*" [PathMaxOccurrence]
+IsRecursive           = "+"
+EBMLAtomName          = ALPHA / DIGIT 0*EBMLNameChar
+EBMLNameChar          = ALPHA / DIGIT / "-" / "."
+
+GlobalPlaceholder     = "(" GlobalParentOccurence "\)"
```
dericed: Since parenthesis are no longer used for recursion, I suggest a cleaner look and use:

`GlobalPlaceholder     = "(" GlobalParentOccurence ")"`

specification.markdown
```diff
@@ -386,41 +386,60 @@ The path defines the allowed storage locations of the EBML Element within an EBM
...
 
-The starting PathDelimiter of the path corresponds to the root of the EBML Document.
+The starting PathDelimiter of EBMLParentPath corresponds to the root of the EBML Document.
 
-In some cases the EBMLLastParent part of the path is an EBMLGlobalParent. A path with a EBMLGlobalParent defines a Global Element; see (#global-elements). Any path that starts with the EBMLFixedParent of the Global Element and matches the occurrences found in the GlobalParentOccurence is a valid path for the Global Element. 
+The `@path` value MUST be unique within the EBML Schema. The `@id` value corresponding to this `@path` MUST NOT be defined for use within another EBML Element with the same EBMLParentPath as this `@path`.
 
-The GlobalParentOccurence part is interpreted as an ABNF Variable Repetition. The repetition amounts correspond to the amount of unspecified Parent Element levels there can be between the EBMLFixedParent and the actual EBMLElementPath.
+A path with a GlobalPlaceholder as the EBMLLastParent defines a Global Element; see (#global-elements).
+If the element has no EBMLLastParent part or the EBMLLastParent part is not a GlobalPlaceholder then the Element is not a Global Element.
```
dericed: Maybe more logical to say `If the element has no EBMLParents part or...`, but still confused.

Suppose you have a Global Master element like:

`/(0-1)/GlobalMaster`

then a Child Element of that would be:
`/(0-1)/GlobalMaster/SubElement`

With this definition, `SubElement` is not considered as a Global Element. Is that intended?

specification.markdown
dericed: I have a slight preference for:

`GlobalPlaceholder = "(" GlobalParentOccurence ")" PathDelimiter`

As in, instead of `\a\(0-1\)global` to have `\a\(0-1)\global`

specification.markdown
```diff
@@ -386,41 +386,60 @@ The path defines the allowed storage locations of the EBML Element within an EBM
...
+If the EBMLElement part contains an IsRecursive part, the EBML Element can occur within itself recursively (see (#recursive)).
 
-The starting PathDelimiter of the path corresponds to the root of the EBML Document.
+The starting PathDelimiter of EBMLParentPath corresponds to the root of the EBML Document.
 
-In some cases the EBMLLastParent part of the path is an EBMLGlobalParent. A path with a EBMLGlobalParent defines a Global Element; see (#global-elements). Any path that starts with the EBMLFixedParent of the Global Element and matches the occurrences found in the GlobalParentOccurence is a valid path for the Global Element. 
+The `@path` value MUST be unique within the EBML Schema. The `@id` value corresponding to this `@path` MUST NOT be defined for use within another EBML Element with the same EBMLParentPath as this `@path`.
 
-The GlobalParentOccurence part is interpreted as an ABNF Variable Repetition. The repetition amounts correspond to the amount of unspecified Parent Element levels there can be between the EBMLFixedParent and the actual EBMLElementPath.
+A path with a GlobalPlaceholder as the EBMLLastParent defines a Global Element; see (#global-elements).
```
dericed: But what about a path with a GlobalPlaceholder as an IntermediatePathAtom?

specification.markdown
```diff
@@ -386,41 +386,60 @@ The path defines the allowed storage locations of the EBML Element within an EBM
...
+The GlobalParentOccurence part is interpreted as the amount of valid EBMLPathAtom parts that can replace the GlobalPlaceholder in the path.
+PathMinOccurrence represents the minimum amount of EBMLPathAtom required to replace the GlobalPlaceholder.
+PathMaxOccurrence represents the maximum amount of EBMLPathAtom possible to replace the GlobalPlaceholder.
 
-PathMaxOccurrence represents the maximum number of element path possible between the EBMLFixedParent and the Global Element EBMLElementPath. It cannot have the value 0 as it would be the Global Element can only be found right after the EBMLFixedParent, in which case it's not a Global Element anymore. If PathMaxOccurrence is not present then there is no upper bound for the permitted number of occurrences of element path possible between the EBMLFixedParent and the Global Element EBMLElementPath.
+If PathMinOccurrence is not present then that GlobalParentOccurence has a PathMinOccurrence value of 0.
+If PathMaxOccurrence is not present then there is no upper bound for the permitted amount of EBMLPathAtom possible to replace the GlobalPlaceholder.
```
dericed: I think we could use a statement, like `PathMaxOccurrence MUST be greater than PathMinOccurrence.`

specification.markdown
```diff
@@ -947,15 +966,15 @@ description: The version of the DocTypeExtension. Different DocTypeExtensionVers
...
+Global Elements are EBML Elements whose EBMLLastParent part of the path has a GlobalPlaceholder. Because it is the last Parent part of the path, a Global Element might also have an EBMLParentPath parts in its path. In this case the Global Element can only be found within this EBMLParentPath path, i.e. it's not fully "global".
 
 A Global Element can be found in many Parent Elements, allowing the same number of occurrences in each Parent where this Element is found.
 
 ### CRC-32 Element
 
 name: CRC-32
 
-path: `*1((1*\)\CRC-32)`
+path: `\(*\)CRC-32`
```
dericed: Needs an update, here an in Void, as we swapped from `*` to `-`.

specification.markdown
robUx4: But that's the point. (*\) means there are plenty of abc\def\ghi\ in place of the placeholder. The \ is "repeated" for each occurence. If you leave it out IMO it's less readable.

specification.markdown
robUx4: (One of) its parent is a global element, but that element in particular is not a global element. The global "placement" only applies to that parent.

specification.markdown
robUx4: That's correct. The SubElement doesn't get to be placed within all kinds of various unknown parents. It has only a fixed parent.

specification.markdown
robUx4: Yeah I left it out because it seems obvious but why not.
It can't be done in the ABNF itself ;(

specification.markdown
robUx4: I removed the `-` change. It slipped in the push but I didn't want it in the end. `-` is part of Element names so it might be confusing an ABNF parser.

robUx4: Never mind. I cannot be confused with a name as it's in parenthesis. It's easier to display a number range with `-` so I put it back.

dericed:
47 commits !_!
should we squash-merge?

dericed:
and if i understand correctly, abnf-examples.txt is simply an example doc for the repository and not part of the RFC, right?

robUx4:
Yes just an example for us. We may put it as an annex if needed.

I prefer to keep the history of commits (and the reasons for the changes) so just rebase and merge.

PR #332, opened 2020-01-04
delete not rendered italic

retokromer:


robUx4:
LGTM. This file is not used though (maybe we should do a folder with old files or simply remove them, they are still in the git history).

retokromer:
okay

PR #337, opened 2020-02-12
fix DocTypeExtension Child Element names

dericed:
fixes https://github.com/cellar-wg/ebml-specification/issues/336


retokromer:
Good catch!

dericed:
Thanks for merging @robUx4. @mcr please advise when is good timing to upload a new version of the draft. It would just have this small fix.

PR #339, opened 2020-05-23
add a more in-depth definition of an enum value

robUx4:
Following the LZO/compression values in Matroska, it seems we are missing an
element to give more information about values in an enum. In this case that
could be a (short) description of the algorithm, a link to a reference
implementation or the standard itself.

It may be too late to add to the spec. But if not that would be a welcome (tiny) addition.

### DIFF
--- a/EBMLSchema.xsd
+++ b/EBMLSchema.xsd
@@ -130,6 +130,7 @@
     </xs:sequence>
     <xs:attribute name="label"/>
     <xs:attribute name="value" use="required"/>
+    <xs:attribute name="definition"/>
   </xs:complexType>
 
   <xs:complexType name="documentationType" mixed="true">
--- a/specification.markdown
+++ b/specification.markdown
@@ -696,6 +696,14 @@ The value represents data that MAY be stored within the EBML Element.
 
 The value attribute is REQUIRED.
 
+#### definition
+
+Within an EBML Schema, the XPath of `@label` attribute is `/EBMLSchema/element/restriction/enum/@definition`.
+
+The definition provides a human readable explanation of the value of the `<enum>` with more information than the concise label attribute.
+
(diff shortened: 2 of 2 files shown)


robUx4:
Alternatively we could reuse the `<documentation>` element in the enum value. Something like this:

```
<enum value="0" label="zlib">
  <documentation purpose="definition">Deflate based compression</documentation>
  <documentation purpose="references">https://zlib.net/</documentation>
</enum>
```

dericed:
I prefer to use <documentation> as this doesn't require any change to ebml and is [already used](https://github.com/cellar-wg/matroska-specification/blob/master/ebml_matroska.xml#L427-L447) in the EBML Schema for Matroska.

robUx4:
(pressed the

> 
> 
> I prefer to use as this doesn't require any change to ebml and is [already used](https://github.com/cellar-wg/matroska-specification/blob/master/ebml_matroska.xml#L427-L447) in the EBML Schema for Matroska.

robUx4:
(pressed the edit button to see you meant `documentation`)

I also prefer it that way.

dericed:
Rather than this I suggest that https://www.ietf.org/id/draft-ietf-cellar-ebml-17.html#section-11.1.6 be changed to say 

```
Within an EBML Schema, the XPath of <documentation> attribute is /EBMLSchema/element/documentation <addthis>or /EBMLSchema/element/enum/documentation</addthis>.

The <documentation> element provides additional information about the EBML Element <addthis>or enumeration value</addthis>. Within the <documentation> element the following XHTML [W3C.SPSD-xhtml-basic-20180327] elements MAY be used: <a>, <br>, <strong>.
```

robUx4:
Closing in favor of #340

PR #340, opened 2020-05-26
Documentation definition fixes

dericed:
This addresses the comment at https://github.com/cellar-wg/ebml-specification/pull/339#issuecomment-634231129 and is intended to supersede https://github.com/cellar-wg/ebml-specification/pull/339 and it fixes a typo where elements of an XML Schema were referred to as attributes.

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -579,9 +579,9 @@ The maxver attribute is OPTIONAL. If the maxver attribute is not present then th
 
 ### \<documentation> Element
 
-Within an EBML Schema, the XPath of `<documentation>` element is `/EBMLSchema/element/documentation`.
+Within an EBML Schema, the XPath of `<documentation>` element are `/EBMLSchema/element/documentation` and `/EBMLSchema/element/restriction/enum/documentation`.
 
-The `<documentation>` element provides additional information about the EBML Element. Within the `<documentation>` element the following XHTML [@!W3C.SPSD-xhtml-basic-20180327] elements MAY be used: `<a>`, `<br>`, `<strong>`.
+The `<documentation>` element provides additional information about EBML Elements or enumeration values. Within the `<documentation>` element the following XHTML [@!W3C.SPSD-xhtml-basic-20180327] elements MAY be used: `<a>`, `<br>`, `<strong>`.
 
 ### \<documentation> Attributes
 
--- a/specification.markdown
+++ b/specification.markdown
@@ -358,7 +358,7 @@ Within an EBML Schema, the XPath of `@ebml` attribute is `/EBMLSchema/@ebml`.
 The ebml attribute is a positive integer that specifies the version of the EBML Header (see (#ebmlversion-element)) used by the EBML Schema. If the attribute is omitted, the EBML Header version is 1.
 
 ### \<element> Element
-  
+
 Within an EBML Schema, the XPath of `<element>` element is `/EBMLSchema/element`.
 
 Each `<element>` defines one EBML Element through the use of several attributes that are defined in (#element-attributes). EBML Schemas MAY contain additional attributes to extend the semantics but MUST NOT conflict with the definitions of the `<element>` attributes defined within this document.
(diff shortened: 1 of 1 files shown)


specification.markdown
```diff
@@ -579,9 +579,9 @@ The maxver attribute is OPTIONAL. If the maxver attribute is not present then th
 
 ### \<documentation> Element
 
-Within an EBML Schema, the XPath of `<documentation>` attribute is `/EBMLSchema/element/documentation`.
+Within an EBML Schema, the XPath of `<documentation>` element are `/EBMLSchema/element/documentation` and `/EBMLSchema/element/restriction/enum/documentation`.
 
-The `<documentation>` element provides additional information about the EBML Element. Within the `<documentation>` element the following XHTML [@!W3C.SPSD-xhtml-basic-20180327] elements MAY be used: `<a>`, `<br>`, `<strong>`.
+The `<documentation>` element provides additional information about EBML Elements or enumeration values. Within the `<documentation>` element the following XHTML [@!W3C.SPSD-xhtml-basic-20180327] elements MAY be used: `<a>`, `<br>`, `<strong>`.
```
retokromer, line 584: Is this still valid? If I remember carefully, _XHTML Basic_ has been superseded by W3C two years ago or so.

dericed: The [xsd](https://github.com/cellar-wg/ebml-specification/blob/master/EBMLSchema.xsd) is valid, the commit resolves the discrepancy between the narrative and the xsd. The reference to XHTML Basic is unchanged by the commit and not in scope of commit, so I suggest it is not blocking, but feel welcome to file an issue.

retokromer: It’s not blocking. I just noticed it while reading this PR. In the meantime I checked, XHTML Basic has been superseded on 27 March 2018.

retokromer: >The reference to XHTML Basic is unchanged by the commit and not in scope of commit, so I suggest it is not blocking, but feel welcome to file an issue.

done in https://github.com/cellar-wg/ebml-specification/issues/341

robUx4 (approved):
LGTM

specification.markdown
```diff
@@ -579,7 +579,7 @@ The maxver attribute is OPTIONAL. If the maxver attribute is not present then th
 
 ### \<documentation> Element
 
-Within an EBML Schema, the XPath of `<documentation>` attribute is `/EBMLSchema/element/documentation`.
+Within an EBML Schema, the XPath of `<documentation>` element is `/EBMLSchema/element/documentation`.
```
mbunkus: Shouldn't this be plural or contain an article? "…the XPath of `<documentation>` elements…" or "…the XPath of the `<documentation>` element…" At the moment it simply sounds grammatically wrong, especially given that the very next sentence does use an article.

Same for the other instances in this commit.

specification.markdown
```diff
@@ -579,9 +579,9 @@ The maxver attribute is OPTIONAL. If the maxver attribute is not present then th
 
 ### \<documentation> Element
 
-Within an EBML Schema, the XPath of `<documentation>` element is `/EBMLSchema/element/documentation`.
+Within an EBML Schema, the XPath of `<documentation>` element are `/EBMLSchema/element/documentation` and `/EBMLSchema/element/restriction/enum/documentation`.
```
mbunkus, line 582: If you're going the plural route, then "XPaths" instead of "XPath", I guess. And again "element" → "elements".

mbunkus (changes requested):

specification.markdown
robUx4: Given there's only one path it seems correct to use singular here.

specification.markdown
robUx4: I agree on this one. If there's an `are` it should be `elements`.

robUx4: Fixes the pluarl

retokromer (approved):

robUx4:
Also corresponds to AUTH48 item 13)

dericed:
LGTM. Thanks for adding that commit @robUx4.

PR #342, opened 2020-06-14
turn "xsd" sourcecode tag into "xml" tag

robUx4:
xsd is not supported and may not produce clean output:
https://www.rfc-editor.org/materials/sourcecode-types.txt

In the end the result is the same.

Fixes item 3.b of the rfc8794 AUTH48 email. It's not really an issue but IMO it's cleaner.


robUx4:
Also in the AUTH48 email issue 14.2)

PR #351, opened 2020-06-14
fix typo

retokromer:
occurence > occurrence


retokromer:
I notice that this is more or less a double of https://github.com/cellar-wg/ebml-specification/pull/346. I close it here but please fix also the other occurrences.

PR #355, opened 2020-06-14
fixed bullet-point text style

robUx4:
As done in AUTH48 17)


dericed:
rebased to fixed conflicts of adjoining lines, the changed line is the same

PR #357, opened 2020-06-27
mark RFC2119 keywords with **<>** to mmark turns them into BCP14 code

robUx4:


specification.markdown
```diff
@@ -219,69 +219,69 @@ Table: Demonstration of VINT_DATA reservation for VINTs of unknown size. {#table
...
+The Master Element contains zero or more other elements. EBML Elements contained within a Master Element **MUST** have the EBMLParentPath of their Element Path equal to the EBMLFullPath of the Master Element Element Path (see (#path)). Element Data stored within Master Elements **SHOULD** only consist of EBML Elements and **SHOULD NOT** contain any data that is not part of an EBML Element. The EBML Schema identifies what Element IDs are valid within the Master Elements for that version of the EBML Document Type. Any data contained within a Master Element that is not part of a Child Element **MUST** be ignored.
 
 ## Binary Element
 
-A Binary Element MUST declare a length in octets from zero to VINTMAX.
+A Binary Element **MUST** declare a length in octets from zero to VINTMAX.
 
 The contents of a Binary Element should not be interpreted by the EBML Reader.
```
retokromer, line 272: Should this not read `**SHOULD NOT**`?

robUx4: It could be but it shouldn't be in this purely formatting patch. You should open an issue to discuss it.

retokromer: okay

robUx4:
Rebased on top of #361 as it should be the first thing merged before we actually change the content of the files.

Makefile
```diff
@@ -14,16 +14,26 @@ all: $(OUTPUT).html $(OUTPUT).txt $(OUTPUT).xml
 	$(info RFC rendering has been tested with mmark version 2.1.1 and xml2rfc 2.30.0, please ensure these are installed and recent enough.)
 
 $(OUTPUT).md: specification.markdown rfc_frontmatter.markdown rfc_backmatter.markdown EBMLSchema.xsd ebml_schema_example.xml
-	cat rfc_frontmatter.markdown $< rfc_backmatter.markdown > $(OUTPUT).md
+	cat rfc_frontmatter.markdown $< rfc_backmatter.markdown | sed "s/@BUILD_DATE@/$(shell date -I)/" > $(OUTPUT).md
```
dericed, line 17: `-I` is an invalid option for the mac version of the `date` utility. Would `date "+%Y-%m-%d"` work for you?

Makefile
```diff
@@ -14,16 +14,26 @@ all: $(OUTPUT).html $(OUTPUT).txt $(OUTPUT).xml
...
 	$(MMARK) $< > $@
 	sed -i -e 's/<sourcecode type="xsd">/<sourcecode type="xml">/' $@
 
-%.html: %.xml
+rfc8794.xml: $(OUTPUT).xml
+	sed -e 's/<?xml version="1.0" encoding="utf-8"?>/<?xml version="1.0" encoding="UTF-8"?>\n<!DOCTYPE rfc SYSTEM "rfc2629-xhtml.ent">/' \
+	-e "s/docName=\"8794\"/docName=\"$(OUTPUT)\"/" \
+	-e 's@<organization></organization>@@' \
+	-e 's@<street></street>@@' \
+	-e 's@BCP 14@BCP\&nbsp;14@' \
```
dericed, line 28: hmm, even with gnu sed, I'm getting 
```
-:71.23: Entity 'nbsp' not defined
described in BCP&nbsp;14 <xref target="RFC2119"></xref> <xref target="RFC8174"><
                      ^
```

Makefile
```diff
@@ -14,16 +14,26 @@ all: $(OUTPUT).html $(OUTPUT).txt $(OUTPUT).xml
...
+	-e 's@<dd><t>@<dd>@' \
+	$< | \
+	awk 1 RS='</t>\n</dd>' ORS='</dd>' | head -n -1 > $@
```
dericed, line 31: Hmm, same issue, works with gnu head but not macOS head.

Makefile
retokromer: FYI: `date +'%F'` (my personal favourite) and `date +"%F"` and `dare +%F` works here.

retokromer: … and on macOS and on Windows Terminal.

robUx4: > Would date "+%Y-%m-%d" work for you?
yes
> date +'%F'
might be nicer if it works everywhere

Makefile
robUx4: I have to find a way to run a macOS shell. This kind of discrepancies can be detected in bootstrap and adjust the commands.

dericed: I don't think that that needs to block the PR. I can see that it works and hope to replace the head and sed with a process that works well natively on mac.

robUx4: OK. I tried with FreeBSD but still have to come up with a proper solution (otherwise a better way to remove the `</t>` from `</t>\n</dd>` than awk would be better). xsltproc is also not an option as it will mess the newlines in the text (I think)

PR #358, opened 2020-06-27
add the first version of the rfc8794 XML source

robUx4:
As provided by RFC editor during the AUTH48 phase.

It doesn't exactly match with the mmark output.


robUx4:
It is worth because ultimately that's the XML that will create the real RFC.

But it also helps to fix the remaining changes (shown in the diff but not in their emails) and know how/if we can generate it properly. Particularly because we're going to generate more documents for Matroska and it's better if they are in the right format from the start.

I'll see about adding some text in the README

PR #360, opened 2020-06-28
replace the custom Matroska reference to one found by xml2rfc

robUx4:
Found in the AUTH48 RFC-editor XML file
https://xml2rfc.tools.ietf.org/public/rfc/bibxml3/reference.I-D.ietf-cellar-matroska.xml


robUx4:
Already merged, in #361, closing.

PR #361, opened 2020-06-28
Generate a rfc8794.xml close to the RFC editor one

robUx4:
This way we can compare and find differences.

It should be the same as https://www.rfc-editor.org/authors/rfc8794-diff.html

The generated file is also kept in the artifacts.

Replaces #359 with a file much closer to the original RFC-editor one thanks to line returns.


robUx4:
We should probably send the rfc8794.xml when we're done with the editing rather than the plain XML from the Markdown source. Otherwise it's harder for the RFC editor to generate their final XML file.

PR #364, opened 2020-06-30
adjust some document references to match the AUTH48 ones

robUx4:
The XML/XPath references are now set locally.

### DIFF
--- a/rfc_backmatter.markdown
+++ b/rfc_backmatter.markdown
@@ -1,42 +1,40 @@
 
 {backmatter}
 
-<reference anchor="ISO.9899.2011">
+<reference anchor="ISO.9899" target="https://www.iso.org/standard/57853.html">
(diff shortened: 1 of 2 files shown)


rfc_backmatter.markdown
```diff
@@ -63,6 +61,60 @@
     <author>
       <organization>The WebM Project</organization>
     </author>
-    <date month="November" year="2017" />
+    <date month="November" year="2017" day="28" />
```
retokromer, line 64: Why not `<date year="2017" month="November" day="28"/>`?

robUx4: No particular reason. In the end mmarkdown reformats it anyway.

retokromer: okay, thanks!

PR #366, opened 2020-06-30
replace some instances of "which" with "that"

robUx4:
To match the AUTH48 text style


robUx4:
Content already merged in other PRs, closing.

PR #368, opened 2020-06-30
replace some instances of plural "semantics" with singular "semantic"

robUx4:
To match the AUTH48 text style

### DIFF
--- a/specification.markdown
+++ b/specification.markdown
@@ -224,7 +224,7 @@ date -- e.g., when its final size isn't known in advance. In
 [@tableVariousSizes], an integer `2` (with a 
 corresponding binary value of 0b10) is shown encoded as different Variable 
 Size Integers with lengths from one octet to four octets. All four encoded
-examples have identical semantics meaning, though the VINT\_WIDTH and the  padding
+examples have identical semantic meaning, though the VINT\_WIDTH and the  padding
 of the VINT\_DATA vary.
 
 Integer | Octet Length | As Represented in VINT (binary)         | As Represented in VINT (hexadecimal)
@@ -776,7 +776,7 @@ The minOccurs attribute is **OPTIONAL**. If the minOccurs
 attribute is not present, then that EBML Element has a minOccurs value of
 0.
 
-The semantics meaning of minOccurs within an EBML Schema is analogous to the meaning of minOccurs within an XML Schema.
+The semantic meaning of minOccurs within an EBML Schema is analogous to the meaning of minOccurs within an XML Schema.
 
 #### maxOccurs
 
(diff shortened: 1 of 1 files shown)


robUx4:
When used as an adjective there's no s (in `semantic meaning`). When used as a noun it [seems the plural form should be used.](https://www.merriam-webster.com/dictionary/semantics)
>noun, plural in form but singular or plural in construction

So maybe `Elements semantic` and `making the semantic invalid` should not be changed ?

Pinging English native speakers (@dericed, @mcr , @SpencerDawkins, @ablwr)

SpencerDawkins:
On Thu, Jul 2, 2020 at 12:45 AM Steve Lhomme <notifications@github.com>
wrote:

> When used as an adjective there's no s (in semantic meaning). When used
> as a noun it seems the plural form should be used.
> <https://www.merriam-webster.com/dictionary/semantics>
>
> noun, plural in form but singular or plural in construction
>
> So maybe Elements semantic and making the semantic invalid should not be
> changed ?
>
> Pinging English native speakers (@dericed <https://github.com/dericed>,
> @mcr <https://github.com/mcr> , @SpencerDawkins
> <https://github.com/SpencerDawkins>, @ablwr <https://github.com/ablwr>)
>
My wife is not entirely sure I'm a native English speaker (but I was born
in Texas).

I reviewed the changes in this PR and find they sound correct to me
("sounding correct" is probably the best approximation for a mongrel
language like English!).

Best,

Spencer

specification.markdown
```diff
@@ -224,7 +224,7 @@ date, e.g. when its final size isn't known in advance. In
 [@tableVariousSizes] an integer `2` (with a 
 corresponding binary value of 0b10) is shown encoded as different Variable 
 Size Integers with lengths from one octet to four octets. All four encoded
-examples have identical semantics meaning though the VINT\_WIDTH and the  padding
+examples have identical semantic meaning though the VINT\_WIDTH and the  padding
```
mcr, line 227: This seems an improvement.

specification.markdown
```diff
@@ -512,7 +512,7 @@ The version of the EBML Header is found in EBMLVersion. An EBML parser can read
 
 The version of the EBML Body is found in EBMLDocTypeVersion. A parser for the particular DocType format can read the EBML Document if it can read either the EBMLDocTypeVersion version of that format or a version equal or higher than the one found in EBMLDocTypeReadVersion.
 
-# Elements semantics
+# Elements semantic
```
mcr, line 515: This feels wrong.  I think that "Element semantics" would be correct.

mbunkus: I concur with "Element semantics" being the way to go here.

robUx4: Fixed

PR #370, opened 2020-06-30
adjust the registry name according to AUTH48

robUx4:


robUx4:
Not sure why one is plural and the other singular. (especially since there will be a lot more IDs than Doctypes)
cc @mcr

PR #374, opened 2020-07-12
Integrate changes from the AUTH48 v2 review

robUx4:


rfc-editor/rfc8794.xml
```diff
@@ -969,17 +891,17 @@ and deprecated by using the <tt>minver</tt> and <tt>maxver</tt> attributes of
 <t>Within an EBML Schema, the XPath of the <tt>@ebml</tt> attribute is <tt>/EBMLSchema/@ebml</tt>.</t>
 <t>The <tt>ebml</tt> attribute is a positive integer that specifies the
 version of the EBML Header (see <xref target="ebmlversion-element"></xref>)
-used by the EBML Schema. If the attribute is omitted, the EBML Header version
+used by the EBML Schema. If the  attribute is omitted, the EBML Header version
```
retokromer, line 894: ?

rfc-editor/rfc8794.xml
```diff
@@ -657,9 +601,9 @@ prevents an Element Data Size from being expressed as the unknown-size
 value. <xref target="tableVintReservation"></xref> clarifies this rule by
 showing a valid and invalid expression of an Element Data Size with a
 VINT_DATA of 127 (which is equal to 2<sup>1*7</sup>-1) and 16,383 (which is equal to
-2<sup>2*7-1</sup>).</t>
+2<sup>2*7</sup>-1).</t>
```
retokromer, line 604: Good catch! I missed that one…

rfc-editor/rfc8794.xml
robUx4: Ah good catch. I'll remove the remaining double spaces.
This XML file is the one the RFC editors untouched, it's not really up for review. But we do need to fix this as it ends up in the `.txt` version of the RFC.

rfc-editor/rfc8794.xml
```diff
@@ -1,32 +1,27 @@
...
-<!-- updated by Chris 02/26/20 -->
-
-
-
-<rfc version="3" ipr="trust200902" docName="draft-ietf-cellar-ebml-17"
-     number="8794" submissionType="IETF" category="std" consensus="true"
-     xml:lang="en" xmlns:xi="http://www.w3.org/2001/XInclude"
-     tocInclude="true" symRefs="true">
+<!-- name="GENERATOR" content="github.com/mmarkdown/mmark Mmark Markdown Processor - mmark.miek.nl" -->
+<rfc version="3" ipr="trust200902" docName="draft-ietf-cellar-ebml-18" submissionType="IETF" category="std" xml:lang="en" xmlns:xi="http://www.w3.org/2001/XInclude" tocInclude="true" symRefs="true" consensus="true" number="8794">
```
retokromer, line 4: Should this not become `draft-ietf-cellar-ebml-19`, like in the Makefile?

rfc-editor/rfc8794.xml
```diff
@@ -214,7 +165,7 @@ length.</dd>
 <section anchor="variable-size-integer"><name>Variable Size Integer</name>
 <t>The Element ID and Element Data Size are both encoded as a Variable Size
 Integer. The Variable Size Integer is composed of a VINT_WIDTH, VINT_MARKER,
-and VINT_DATA, in that order. Variable Size Integers <bcp14>MUST</bcp14>
+and VINT_DATA, in that order. Variable Size Integers  <bcp14>MUST</bcp14>
```
retokromer, line 168: double space

rfc-editor/rfc8794.xml
```diff
@@ -299,14 +250,14 @@ depicted by <tt>0</tt>, the VINT_MARKER as <tt>1</tt>, and the VINT_DATA as
...
 </table><t>A Variable Size Integer may be rendered at octet lengths larger
 than needed to store the data in order to facilitate overwriting it at a later
-date -- e.g., when its final size isn't known in advance. In <xref
-target="tableVariousSizes"></xref>, an integer <tt>2</tt> (with a
+date -- e.g., when its final size isn't known in advance. In
+<xref target="tableVariousSizes"></xref>, an integer <tt>2</tt> (with a
 corresponding binary value of 0b10) is shown encoded as different Variable
 Size Integers with lengths from one octet to four octets. All four encoded
-examples have identical semantic meaning, though the VINT_WIDTH and the padding
+examples have identical semantic meaning, though the VINT_WIDTH and the  padding
```
retokromer, line 257: double space

rfc-editor/rfc8794.xml
```diff
@@ -1823,19 +1623,19 @@ then the first instance of the Identically Recurring Element should be used
...
 expressed as Hexadecimal Floating-Point Constants as defined in the C11
-standard <xref target="ISO.9899"></xref> (see Section 6.4.4.2 on Floating
+standard <xref target="ISO.9899"></xref> (see section 6.4.4.2 on Floating
 Constants). <xref target="tableFloatExamples"></xref> provides examples of
-expressions of Float ranges.</t>
-<table anchor="tableFloatExamples"><name>Example of Floating-Point Values and
-Ranges as Decimal and Hexadecimal Floating-Point Constants</name>
+expressions of float ranges.</t>
+<table anchor="tableFloatExamples"><name>Example of Floating-Point values and
+ranges as decimal and Hexadecimal Floating-Point Constants </name>
```
retokromer, line 1634: Is there a reason that `Floating-Point` and `Floating-Point Constants` are still “capitalised”?

rfc-editor/rfc8794.xml
```diff
@@ -2891,16 +2589,44 @@ Original:
     </author>
     <date year="1994"></date>
   </front>
-     <seriesInfo name="ITU-T Recommendation" value="V.42" />
+  <seriesInfo name="ITU-T Recommendation" value="V.42"></seriesInfo>
+</reference>
+<reference anchor="XML" target="http://www.w3.org/TR/2008/REC-xml-20081126/">
```
retokromer, line 2594: `https://`

rfc-editor/rfc8794.xml
```diff
@@ -2909,82 +2635,45 @@ be appropriate to change the reference to the newer version?
     </author>
     <date year="2011"></date>
   </front>
-<seriesInfo name="ISO/IEC" value="9899:2011" />
+  <seriesInfo name="ISO/IEC" value="9899:2011"></seriesInfo>
```
retokromer, line 2638: This ISO has been reviewed in 2018.

rfc-editor/rfc8794.xml
```diff
@@ -2909,82 +2635,45 @@ be appropriate to change the reference to the newer version?
...
   <front>
-    <title>IEEE Standard for Binary Floating-Point Arithmetic</title>
+    <title>Data communication -- High-level data link control procedures -- Frame structure</title>
     <author>
-      <organization>IEEE</organization>
+      <organization>International Organization for Standardization</organization>
     </author>
+    <date year="1979"></date>
   </front>
+  <seriesInfo name="ISO" value="Standard 3309"></seriesInfo>
```
retokromer, line 2661: I guess, this has been superseded by ISO/IEC 13239:2002.

rfc-editor/rfc8794.xml
robUx4: No, that's the version we sent them and they edited. Our version will be `draft-ietf-cellar-ebml-19` but it's not generated in the same location.

rfc-editor/rfc8794.xml
robUx4: As I said in the email, [C18 is pretty much C11 with a few fixes](https://en.wikipedia.org/wiki/C18_(C_standard_revision)). People usually refer to C11 and not C18. So I prefer to keep 2011.

rfc-editor/rfc8794.xml
robUx4: I think it refers to `expressed as Hexadecimal Floating-Point Constants as defined in the C11 standard [@!ISO.9899]` which is considered a defined term.

rfc-editor/rfc8794.xml
robUx4: Indeed

rfc-editor/rfc8794.xml
robUx4: I looked at all RFCs (didn't read them all) and there's no mention of ISO 13239. There are plenty of ISO 3309, so I would rather keep in line. The date and format differ from what we use so I will edit that.

rfc-editor/rfc8794.xml
retokromer: okay, thank you for the explanation

rfc-editor/rfc8794.xml
retokromer: no problem at all, it was just to have current references in the text (I personally use often an even older C myself ;-)

rfc-editor/rfc8794.xml
retokromer: okay, thank you for the explanation

robUx4: Actually, looking at the ISO links we use, only the latest version is available for download (after some $ exchange). So we link to something people can't even download. They can still follow links. For example rfc8122 uses https://www.iso.org/standard/64854.html which is not the latest and can't be downloaded.

I still prefer to use the most common names for standards (and also avoid ISO if I can).

retokromer: I agree plenty with you, even having worked many years in ISO TCs. Nobody is perfect…

robUx4: I created a separate PR #383 to clean all the references.

robUx4:
Merged via #384

PR #379, opened 2020-07-12
Clarify the 0x is for notation, not storing

robUx4:
Different text than suggested by the RFC editors.


dericed:
I merged a few of the other minor recent PR's from auth48. Can you fix the conflict in this one?

PR #383, opened 2020-07-13
Clean external references

robUx4:
Fixed some links, used more consistent dates, reordered, cleaned some text.

Will need rebase on top of #374


dericed:
ping to @mbunkus on this and https://github.com/cellar-wg/ebml-specification/pull/374

PR #384, opened 2020-07-13
Auth48 v2b

dericed:
Hi @robUx4, https://github.com/cellar-wg/ebml-specification/pull/374 had merge conflicts so I rebased those commits from master to this branch.


robUx4:
OK, next time push force it on top of `auth48-v2` so we don't end up with 2 PRs with the same code. (I verified my rebase edits match yours).

PR #387, opened 2020-07-17
Use a more stable and actual URL for namespace of our EBML Schema

robUx4:


mbunkus (approved):

retokromer (approved):

mbunkus (changes requested):
Steve, please see Richard's reply on the CELLAR ML regarding preference of a URN over URL, and if URL which one to use.

dericed:
I suggest that we also include a declaration of the namespace in the narrative of the RFC, equivalent of https://tools.ietf.org/html/rfc4826#section-3.4.4.

dericed:
I also prefer `urn:ietf:rfc:8794` as well, particularly after reviewing @mcr's comments.

robUx4:
Closing this in favor of #388

PR #388, opened 2020-07-17
clarify EBMLSchema namespace

dericed:
alternative to https://github.com/cellar-wg/ebml-specification/pull/387


robUx4:
About the formatting, I think the ``` is overkill here. Simple backticks should be enough.
If that's OK I'll update this PR before I merge it.

dericed:
changed to a simple backtick quote

robUx4:
Turned it into a regular sentence. Otherwise it's just quoted text with missing punctuation.

dericed:
Lgtm

PR #393, opened 2021-01-18
EBMLMaxIDLength should be greater or equal 4

StefH:
https://github.com/ietf-wg-cellar/ebml-specification/blob/master/specification.markdown#ebmlmaxidlength-element


robUx4:
Ah didn't see it before I pushed the fixed. Also I think it's better to just escape the >. Having a special character for >= may not result in correct RFC text.

PR #396, opened 2021-02-08
fix maxOccurs default value in the EBML Schema

### LINKED ISSUE #395: Potentially incorrect default value for maxOccurs in XSD
In `EBMLSchema.xsd` the default value for `maxOccurs` is defined as `1`.

The specification says the following about `maxOccurs`:

> [...] **If the `maxOccurs` attribute is not present, then there is no upper bound** for the permitted number of occurrences of this EBML Element within its Parent Element or within the EBML Document, depending on whether or not the EBMLParentPath of the EBML Element is empty.
>
> The semantic meaning of maxOccurs within an EBML Schema is analogous to the meaning of maxOccurs within an XML Schema; **when it is not present, it's similar to xml:maxOccurs="unbounded" in an XML Schema.**

In my opinion the XSD definition could be misleading. I propose to define `maxOccurs` as a union type or define `0` as a special value.

In case `maxOccurs` is defined as a union type:

- Schema change
  ```
    <xs:attribute name="maxOccurs" default="unbounded">
      <xs:simpleType>
        <xs:union> 
          <xs:simpleType> 
            <xs:restriction base="xs:integer">
              <xs:minInclusive value="0"/>
            </xs:restriction>
          </xs:simpleType> 
          <xs:simpleType> 
            <xs:restriction base="xs:string"> 
              <xs:enumeration value="unbounded"/> 
            </xs:restriction> 
          </xs:simpleType> 
        </xs:union>
      </xs:simpleType>
    </xs:attribute>
  ```
- This change would let `maxOccurs` behave almost the same as the attribute with the same name in XML Schema.

In case `0` is defined as a special value:

- Schema change
  ```
    <xs:attribute name="maxOccurs" default="0">
      <xs:simpleType>
        <xs:restriction base="xs:integer">
          <xs:minInclusive value="0"/>
        </xs:restriction>
      </xs:simpleType>
    </xs:attribute>
  ```
- The zero value has to be defined as a special value in the specification. From a logical perspective this could make sense because limiting the maximal occurrence to 0 is an invalid constraint.

nerg4l:
A PR as promised.

Fixes #395

Feel free to add suggestion for the wording, or simply commit your changes on this PR (I allowed edits by maintainers).

### DIFF
--- a/EBMLSchema.xsd
+++ b/EBMLSchema.xsd
@@ -61,11 +61,20 @@
         </xs:restriction>
       </xs:simpleType>
     </xs:attribute>
-    <xs:attribute name="maxOccurs" default="1">
+    <xs:attribute name="maxOccurs" default="unbounded">
       <xs:simpleType>
-        <xs:restriction base="xs:integer">
-          <xs:minInclusive value="0"/>
-        </xs:restriction>
+        <xs:union>
+          <xs:simpleType>
+            <xs:restriction base="xs:integer">
+              <xs:minInclusive value="0"/>
+            </xs:restriction>
+          </xs:simpleType>
+          <xs:simpleType>
+            <xs:restriction base="xs:string">
+              <xs:enumeration value="unbounded"/>
+            </xs:restriction>
+          </xs:simpleType>
+        </xs:union>
       </xs:simpleType>
     </xs:attribute>
     <xs:attribute name="range"/>
--- a/specification.markdown
+++ b/specification.markdown
@@ -789,8 +789,9 @@ The semantic meaning of `minOccurs` within an EBML Schema is analogous to the me
 Within an EBML Schema, the XPath of the `@maxOccurs` attribute is
 `/EBMLSchema/element/@maxOccurs`.
 
-`maxOccurs` is a nonnegative integer expressing the maximum permitted number
-of occurrences of this EBML Element within its Parent Element.
+`maxOccurs` is a nonnegative integer, or the term "unbounded" expressing
+the maximum permitted number of occurrences of this EBML Element within its
+Parent Element.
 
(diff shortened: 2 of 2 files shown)


robUx4:
I think you should split the PR in 2. While the XSD fix is likely to make it, I'm less confident that a change of text will occur.

robUx4:
The Matroska XML file still validates with the XSD change (phew).

nerg4l:
Is it okay if I leave the change in the specification to you? In my opinion the change is needed because the following statement is now incorrect:

> `maxOccurs` is a nonnegative integer expressing the maximum permitted number
> of occurrences of this EBML Element within its Parent Element.

The rest can be kept as is. I just wanted to clarify the meaning of "unbounded".

robUx4 (approved):
LGTM

robUx4:
@mcr I sent an email to the mailing list so this can be handled through the proper channels
https://mailarchive.ietf.org/arch/msg/cellar/SwUeIWgHsf0ZZ8xDu6ehGT0ZcII/

robUx4:
> 
> 
> Is it okay if I leave the change in the specification to you? In my opinion the change is needed because the following statement is now incorrect:
> 
> > `maxOccurs` is a nonnegative integer expressing the maximum permitted number
> > of occurrences of this EBML Element within its Parent Element.
> 
> The rest can be kept as is. I just wanted to clarify the meaning of "unbounded".

Just create a new PR with this other patch. It might be approved but I don't know how/if it will end up in the ERRATA.

nerg4l:
Okay, thanks for the clarification

retokromer (approved):

robUx4:
Merge thanks !
Now we have to do the RFC errata.

PR #402, opened 2021-03-25
Void element: remove the maxOccurs attribute

hubblec4:
The Void element can be used multiple.


JeromeMartinez (approved):

retokromer (approved):

robUx4:
Very good catch!
This should also be an errata of the RFC.

This file can be useful for tools using the XML Schema and merging this file with their own format Schema (we don't do it for our Matroska tools). That would create false checks on the Void elements.

robUx4:
Removed the errata label as ebml.xml is not in the RFC

PR #406, opened 2021-04-04
add minver/maxver attributes to enum values

robUx4:
This allows adding new enum values without forcing old readers to support them.

The explanation is copied from the minver attribute of "element"

This is necessary (IMO) to add new values like [zstd compression](https://github.com/ietf-wg-cellar/matroska-specification/pull/423)

### DIFF
--- a/EBMLSchema.xsd
+++ b/EBMLSchema.xsd
@@ -146,11 +146,12 @@
     </xs:sequence>
     <xs:attribute name="label"/>
     <xs:attribute name="value" use="required"/>
+    <xs:attribute name="minver"/>
   </xs:complexType>
 
   <xs:complexType name="documentationType" mixed="true">
     <xs:sequence>
-      <xs:element name="a"      type="xhtml:xhtml.a.type" 
+      <xs:element name="a"      type="xhtml:xhtml.a.type"
         minOccurs="0" maxOccurs="unbounded"/>
       <xs:element name="br"     type="xhtml:xhtml.br.type"
         minOccurs="0" maxOccurs="unbounded"/>
--- a/EBMLSchema.xsd
+++ b/EBMLSchema.xsd
@@ -147,6 +147,7 @@
     <xs:attribute name="label"/>
     <xs:attribute name="value" use="required"/>
     <xs:attribute name="minver"/>
+    <xs:attribute name="maxver"/>
   </xs:complexType>
 
   <xs:complexType name="documentationType" mixed="true">
--- a/update1/update1.markdown
+++ b/update1/update1.markdown
@@ -87,6 +87,23 @@ The `minver` attribute is **OPTIONAL**. If the `minver`
 attribute is not present, the enum's minimum version is the same as the
 EBML Element's minimum version this enum belongs to.
 
+## maxver Attribute
+
+A new attribute is added to the list of `<enum>` attributes found in [@?RFC8794, section 11.1.13].
+
+Within an EBML Schema, the XPath of the `@value` attribute is
+`/EBMLSchema/element/restriction/enum/@maxver`.
+
+The `maxver` (maximum version) attribute stores a nonnegative integer that
+represents the last or most recent version of the docType to support
+the enum value. `maxver` **MUST** be greater than or equal to
+the first version of the docType to support the enum value, whether it's set
+in the `minver` attribute of the enum or implied.
+
+The `maxver` attribute is **OPTIONAL**. If the `maxver`
+attribute is not present, the enum's maximum version is the same as the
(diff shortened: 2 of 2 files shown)


robUx4:
Actually the default value should not be 1 but the `minver` value of the element...

hubblec4:
Thats the same thing with the element attribute "minver" and "maxver"

What is when this enum is removed in the next ebml-version... and what is when this enum returns in a next next version?

robUx4:
Indeed, there's no reason to have one and not the other. I added both.

specification.markdown
```diff
@@ -1155,6 +1155,32 @@ The value represents data that **MAY** be stored within the EBML Element.
...
+attribute is not present, then the enum value has the same value as the `minver` attribute of the
+EBML Element it belongs to.
+
+#### maxver
+
+Within an EBML Schema, the XPath of the `@value` attribute is
+`/EBMLSchema/element/restriction/enum/@maxver`.
+
+The `maxver` (maximum version) attribute stores a nonnegative integer that
+represents the last or most recent version of the docType to support the
```
hubblec4, line 1177: "the" is twice there

specification.markdown
```diff
@@ -1155,6 +1155,32 @@ The value represents data that **MAY** be stored within the EBML Element.
...
+EBML Element it belongs to.
+
+#### maxver
+
+Within an EBML Schema, the XPath of the `@value` attribute is
+`/EBMLSchema/element/restriction/enum/@maxver`.
+
+The `maxver` (maximum version) attribute stores a nonnegative integer that
+represents the last or most recent version of the docType to support
+the enum value. `maxver` **MUST** be greater than or equal to the `minver` attribute.
```
mbunkus, line 1178: Doesn't this imply that for the `maxver` attribute to exist, the `minver` attribute must exist as well? Isn't the following more what you'd want to express:

* Enum's `maxver` must not be less than enum's `minver` if that exists
* Enum's `maxver` must not be less than element's `minver` if that exists

specification.markdown
```diff
@@ -1155,6 +1155,32 @@ The value represents data that **MAY** be stored within the EBML Element.
...
+
+The `maxver` attribute is **OPTIONAL**. If the `maxver` attribute is
+not present, then the enum value has a maximum version equal to the value
+stored in the `version` attribute of `<EBMLSchema>`.
```
mbunkus, line 1182: Same concern about the wording as above. Maybe:

> If the `maxver` attribute is not present, then the enum value's maximum version is the value stored in the `version` attribute of `<EBMLSchema>`.

mbunkus (changes requested):

specification.markdown
```diff
@@ -1155,6 +1155,32 @@ The value represents data that **MAY** be stored within the EBML Element.
...
 The `value` attribute is **REQUIRED**.
 
+#### minver
+
+Within an EBML Schema, the XPath of the `@value` attribute is
+`/EBMLSchema/element/restriction/enum/@minver`.
+
+The `minver` (minimum version) attribute stores a nonnegative integer that
+represents the first version of the docType to support the enum value. The value
+**MUST** be greater than or equal to the `minver` attribute value of the EBML Element it belongs to.
```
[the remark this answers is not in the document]

specification.markdown
```diff
@@ -1155,6 +1155,32 @@ The value represents data that **MAY** be stored within the EBML Element.
...
+
+The `minver` attribute is **OPTIONAL**. If the `minver`
+attribute is not present, then the enum value has the same value as the `minver` attribute of the
```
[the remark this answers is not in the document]
robUx4: Yeah, it wouldn't make sens that it "**MUST** be greater than or equal to the EBML Element's minimum version this enum belongs to" and as the same time have a default value that is not.

specification.markdown

specification.markdown
robUx4: This text is mostly the same as the `maxver` of the `<element>`, which is also not very precise. The idea is that it's greater than or equal to the enum minimum version, whether it is written or assumed.

robUx4 (commented):
Updated according to comments.

hubblec4 (approved):

hubblec4:
@robUx4 can you change the title to
"add a minver and a maxver attribute to enum values"

specification.markdown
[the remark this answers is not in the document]
robUx4: looks like I did the changes in the wrong place

retokromer (approved):

robUx4:
Merging so we can publish the first draft and discuss changes if needed.

PR #413, opened 2021-06-20
generate a new document for corrections

### LINKED ISSUE #412: create a new document containing all errata and additions to RFC 8794
As discussed during our [last IETF meeting](https://codimd.ietf.org/notes-cellar-virtual), this should go in a new document which not an errata document nor a replacing RFC.

We need to collect all the [PRs with errata label](https://github.com/ietf-wg-cellar/ebml-specification/pulls?q=is%3Apr+label%3Aerrata-rfc8794) and put them in a new generated document.

robUx4:
The format is simpler that the one in [RFC 8540](https://www.rfc-editor.org/rfc/rfc8540.html). If needed I can update the format.

We may also add `ebml.xml` as an annex, making it normative. But it's more work than necessary, possibly leading to new fixes...

Fixes #412

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
-VERSION := 21
-STATUS := draft-
-OUTPUT := $(STATUS)ietf-cellar-ebml-$(VERSION)
+VERSION_EBML := 21
+STATUS_EBML := draft-
+OUTPUT_EBML := $(STATUS_EBML)ietf-cellar-ebml-$(VERSION_EBML)
+VERSION_CORRECTION := 01
+STATUS_CORRECTION := draft-
+OUTPUT_CORRECTION := $(STATUS_EBML)ietf-cellar-rfc8794-corrections-$(VERSION_CORRECTION)
 
 XML2RFC_CALL := xml2rfc
 MMARK_CALL := mmark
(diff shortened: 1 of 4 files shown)


robUx4:
We need to create a new entry in the CELLAR documents https://datatracker.ietf.org/wg/cellar/documents/

SpencerDawkins:
@robUx4, thanks for starting this work. I'm not good at generating the EBML specifications, but I did poke around a bit. 

If I could make a couple of suggestions (which you may have already taken care of - my apologies, if so):

- I'd suggest using draft-ietf-cellar-rfc8794-corrections as the output file name. Unless @mcr has "other thoughts" ("spencer-corrections"?), I don't see that we need to go through the "post as individual and see if the working group adopts it, and then rename" shuffle. 
- It is going to be much easier to take this draft through the approvals process, at some point in time, if you actually cut-and-paste the text being modified, and then the complete text with corrections, for each correction, using "OLD" and "NEW" to prefix them, just as we would for errata (if we were doing the RFC errata process). If you're correcting a sentence in the middle of a paragraph, show the entire paragraph (OLD and NEW), so that it's obvious to the reader where this text is in the document. 
- If you can summarize in a sentence or two about why the original text needs to be corrected, that would be very helpful. I see something like that (I think!), so thank you for that. 
- Just looking at the corrections listed so far - I see that https://www.rfc-editor.org/rfc/rfc8141.html **obsoletes** RFC 2141, but I'm not seeing that RFC 2141 has been **deprecated** (unless I'm missing it). Is the goal here to make use of extensions that are allowed in RFC 8141 (https://www.rfc-editor.org/rfc/rfc8141.html#appendix-B),, or just to use the current version of the standards-track URN specification? Either way, we should talk about who deprecated RFC 2141, or change the wording to Obsolete.

robUx4:
* done
* OK, alhtough for the XML Schema it may look a bit odd
* will do
* #389 mentions deprecated but it's just obsolete. maybe we don't really need this change ?

robUx4:
Given how RFC errata work, I opted to keep updating the current document with changes, so it can be diff'ed with the current RFC to produce the errata text.

robUx4:
There is now an [rfc8794 branch](https://github.com/ietf-wg-cellar/ebml-specification/tree/rfc8794) that collects all the errata so we can regenerate a plain with these errata applies.

robUx4:
This should probably be integrated into the update document that adds new attributes to EBML definitions.

PR #414, opened 2021-06-20
Range and number of valid element IDs

hubblec4:
After merging #408 the range and number of valid element IDs with 1 octet no longer correct.


robUx4:
OK with the change although it will need to be mentioned in the new document from #413

PR #416, opened 2021-08-26
fix Matroska draft reference

robUx4:
`I-D.ietf-cellar-matroska` on an unpublished document doesn't work anymore.

I-D.draft-ietf-cellar-matroska or I-D.draft-ietf-cellar-matroska-07 should work
but they don't, xml2rfc doesn't find load the XML document that exists.

So we manually set a matching reference in the list of references.

### DIFF
--- a/rfc_backmatter.markdown
+++ b/rfc_backmatter.markdown
@@ -44,6 +44,18 @@
   <seriesInfo name="ITU-T Recommendation" value="V.42" />
 </reference>
 
+<reference anchor="I-D.ietf-cellar-matroska">
+  <front>
+    <title>Matroska Media Container Format Specifications</title>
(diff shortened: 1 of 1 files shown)


robUx4:
Updated with a link to version 8 of the Matroska draft.

retokromer:
This needs to be updated again …

You could use the following link, which at least always points to the last available version:
```
<format type="TXT" target="https://datatracker.ietf.org/doc/draft-ietf-cellar-matroska//>
```

robUx4:
Actually not changing anything works automagically. xml2rfc picks the latest (v10) from the `I-D.ietf-cellar-matroska` reference. It might have been a bug when I did the "fix" or it was in an "unpublished" our outdated state at that time.

PR #418, opened 2021-11-21
EBML Schema path: add grouping to EBMLAtomName

matthewleon:
The `EBMLAtomName` ABNF rule was incorrectly specified: It failed to account for the higher operator precedence of concatenation vs. alternative, as stated in RFC 5234 section 3.10: https://datatracker.ietf.org/doc/html/rfc5234#section-3.10


robUx4:
Indeed, well spotted.

Merging, it should go in some errata we will publish or an update of the RFC.

PR #419, opened 2021-12-04
EBML Schema XML: adjust reference to 'mastroska' DocType

matthewleon:
A fix to some copy-pasta. The previous language here seems inappropriate for a generic EBML Schema.


robUx4:
Indeed, `our case` in the context of the EBML spec doesn't make much sense.

retokromer (approved):

PR #431, opened 2024-09-16
Fix expression for element size's maximum value

mjbshaw:


JeromeMartinez (approved):

retokromer (approved):

robUx4:
This does not render the exponent correctly when run through mmark and xml2rfc.

Before the patch in [RFC 8794 Section 6.3](https://www.rfc-editor.org/rfc/rfc8794.html#section-6.3) we have: 
> with an octet length of 8 is able to express a size of 2<sup>(56)</sup>-2 or 72,057,594,037,927,934 octets (or about 72 petabytes)

When this patch is applied we have:
> with an octet length of 8 is able to express a size of 2^(56)-2 or 72,057,594,037,927,934 octets (or about 72 petabytes)

The `<sup>` is gone.

This would actually be a revert of a78b551f9db657f67a8303247b201678832feb54 which fixed this.

mjbshaw:
Thanks for the review. Sorry, I didn't realize this was [an extension in mmark](https://mmark.miek.nl/post/syntax/#super--and-subscript).

PR #433, opened 2026-04-26
Add a new document with the "added" attribute

robUx4:
So we can keep using the same `minver` version and distinguish elements between version of a specficiations.

We could also name it "revision" and use incremental values between published versions.
It carries less information than `added` but allows clean distinction of document/spec updates.

### DIFF
--- a/EBMLSchema.xsd
+++ b/EBMLSchema.xsd
@@ -114,6 +114,13 @@
         </xs:restriction>
       </xs:simpleType>
     </xs:attribute>
+    <xs:attribute name="added" default="0">
+      <xs:simpleType>
+        <xs:restriction base="xs:integer">
+          <xs:minInclusive value="0"/>
+        </xs:restriction>
+      </xs:simpleType>
+    </xs:attribute>
   </xs:complexType>
 
   <xs:complexType name="restrictionType">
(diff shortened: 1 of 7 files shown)


added.markdown
```diff
@@ -0,0 +1,59 @@
...
+
+The key words "**MUST**", "**MUST NOT**",
+"**REQUIRED**", "**SHALL**", "**SHALL NOT**",
+"**SHOULD**", "**SHOULD NOT**",
+"**RECOMMENDED**", "**NOT RECOMMENDED**",
+"**MAY**", and "**OPTIONAL**" in this document are to be interpreted as
+described in BCP 14 [@!RFC2119] [@!RFC8174] 
+when, and only when, they appear in all capitals, as shown here.
+
+# added Attribute
```
retokromer: nit: I would also put a capital letter to "Added".

retokromer (commented):
LGTM

added.markdown
robUx4: In [RFC8794](https://www.rfc-editor.org/rfc/rfc8794.html#name-ebmlschema-attributes) the attribute names are kept with the same case as they are written in the XML. But I could add a new section `<EBMLSchema> New Attributes` to make it more similar.

retokromer: Typographically speaking, it’s a bit strange, but I understand why it is that way.

robUx4:
Merging so I can update #406 to be added to this document

